PR #133, opened 2015-08-21
Feature/347 :: Load Average Config. Option

jaswsinc:
One commit (https://github.com/websharks/zencache-pro/commit/e767afb823c96e56aa37c611d9c0341707329978) resolves https://github.com/websharks/zencache/issues/347 (pro version only).

---


jaswsinc:
@raamdev I just discovered a bug in this latest update. Fixed in https://github.com/websharks/zencache-pro/commit/39a65d5fcd8cabb3b51f78bd02c9687b290bcf71

raamdev:
@jaswsinc Regarding the note about `sys_getloadavg()` not being available on Windows: Does that mean the "disable cache expiration if the server load is high" feature will not work at all on Windows? If so, the field to enter the load average should not be usable. Also, the description of the feature should be greyed out.

jaswsinc:
> If so, the field to enter the load average should not be usable. Also, the description of the feature should be greyed out.

Correct. Agree.

jaswsinc:
FYI: This branch is running at websharks-inc.com, s2member.com, and zencache.com now.

jaswsinc:
and comment-mail.com

raamdev:
> Correct. Agree.

Besides greying out the description and the text box for this feature in the UI, where else should we disable this feature when running on Windows? Should I add an `&& stripos(PHP_OS, 'win') !== 0` to this conditional? https://github.com/websharks/zencache-pro/blob/e767afb823c96e56aa37c611d9c0341707329978/src/includes/closures/Ac/ObUtils.php#L191

jaswsinc:
> Should I add an && stripos(PHP_OS, 'win') !== 0 to this conditional? https://github.com/websharks/zencache-pro/blob/e767afb823c96e56aa37c611d9c0341707329978/src/includes/closures/Ac/ObUtils.php#L191

You could, yes. However, I believe we have this covered already here. I was shooting for feature detection instead of OS detection if we can; i.e., to check if this function is available instead of checking for the OS. See: https://github.com/websharks/zencache-pro/blob/feature/347/src/includes/closures/Shared/SysUtils.php#L15

**Action Required:** My feeling is that we should do the same in the alert. This check could also include a call to `$this->plugin->sysLoadAverages()` to see if it's empty first, and then check if it is Windows. See: https://github.com/websharks/zencache-pro/blob/feature/347/src/includes/classes/MenuPageOptions.php#L448

As for OS-detection in the rest of the codebase, that should be covered by the fact that `sysLoadAverages()` returns an empty array whenever it is not possible, and all calls to this method anticipate this being the case on some servers; i.e., we only deal with load averages if we have them.

raamdev:
> Action Required: My feeling is that we should do the same in the alert. This check could also include a call to $this->plugin->sysLoadAverages() to see if it's empty first, and then check if it is Windows.

If we check if `!$this->plugin->sysLoadAverages()`, that may also indicate that `sys_getloadavg()` has been disabled or is unavailable on that server, not necessarily that the server is running Windows. In that case, the message that's displayed on the UI should also mention, "or `sys_getloadavg()` is not available or has been disabled on your server.", correct?

jaswsinc:
> If we check if !$this->plugin->sysLoadAverages(), that may also indicate that sys_getloadavg() has been disabled or is unavailable on that server, not necessarily that the server is running Windows.

Exactly.

> In that case, the message that's displayed on the UI should also mention, "or sys_getloadavg() is not available or has been disabled on your server.", correct?

Agree. We could also check for Windows and improve the message by being more OS-specific in the notice that we issue to a site owner, but I think the idea of leaving is generalized is absolute OK also.

raamdev:
## See https://github.com/websharks/zencache-pro/commit/a3386e7bb1bdc5ddf0ccddf3b27c10b8c2742f1b

@jaswsinc Quick review here, please?

## Windows 

`!$this->plugin->sysLoadAverages() && stripos(PHP_OS, 'win') === 0`



## `sys_getloadavg()` not available

`!$this->plugin->sysLoadAverages() && stripos(PHP_OS, 'win') !== 0`

raamdev:
@jaswsinc  Done: https://github.com/websharks/zencache-pro/commit/32e36c66ae51336659c7db9d543f156a53ba73ad

raamdev:
@jaswsinc Maybe the Pro Preview should prevent any of the description text from being greyed out?

raamdev:
In fact we could disable the check altogether when doing a Pro Preview:

```php
$_sys_getloadavg_unavailable = ($this->plugin->isProPreview() ? FALSE : !$this->plugin->sysLoadAverages());
```

jaswsinc:
> In fact we could disable the check altogether when doing a Pro Preview:

Cool. Yeah, that should work.

raamdev:
@jaswsinc Awesome. Thanks for the feedback! :-) Merged!

jaswsinc:
Yay! Thanks for improving that.

PR #136, opened 2015-08-28
The second attempt at adding feature request: ZenCache Version Number…

renzms:
…, please see: websharks/zencache#502 for original commit


jaswsinc:
@renzms TY. See my notes here regarding changes in this branch:
https://github.com/websharks/zencache/issues/502#issuecomment-135763648

raamdev:
Looking great, @renzms! :-) I just tested this locally and it appears to be working as expected. 

Could you please change `ZenCache™` to `ZenCache™ Pro`? See my screenshots here: https://github.com/websharks/zencache/issues/502#issuecomment-128787903

renzms:
@raamdev ,Thanks for the reminder, must've forgotten it. All good now!
pro plugin updater renz s test area wordpress

raamdev:
@renzms Nice! Thanks!

@jaswsinc This looks ready to merge to me. Since you were helping Renz with a few things on this PR, would you mind giving it a quick review to confirm that it's ready to merge?

jaswsinc:
@renzms Wow, that looks fantastic!

@raamdev writes...

> This looks ready to merge to me. Since you were helping Renz with a few things on this PR, would you mind giving it a quick review to confirm that it's ready to merge?

What's left here is to implement an API call that checks for the latest Lite version also, and a conditional that differentiates between the current instance of ZenCache being the Lite or Pro version; i.e., which label and version to display. So I would hold off on merging this for the moment. There is one more step to complete. I'll try to outline that for @renzms later today.

raamdev:
> What's left here is to implement an API call that checks for the latest Lite version also

I assume you're referring to Lite+?

jaswsinc:
Hmm. Lite+. I see what you mean. Well, I was working from your comment [here](https://github.com/websharks/zencache/issues/502#issuecomment-128787903), which included this for Lite also. Now that you mention it though, given that we will have two different Lite variations, maybe it is better if we only do this for the Pro version for now, and come back to this later?

 If so, we just need to wrap [this block](https://github.com/websharks/zencache-pro/blob/c0b48024f7687cb92a2c2e1cd8ed61483fddecf2/src/includes/classes/MenuPageOptions.php#L59-L68) with an `if(IS_PRO){}` check. If that's something that you can do, then I'd say go ahead and merge this and then make that tweak.

raamdev:
> Well, I was working from your comment here, which included this for Lite also.

I guess you missed my follow-up comment here: https://github.com/websharks/zencache/issues/502#issuecomment-135555295 :-)

jaswsinc:
I saw that comment that you made, but I got confused shortly after. After you made that comment (agreeing with me), I then realized that you had originally mentioned Lite being implemented also, from your earlier screenshots.

So I thought perhaps that you meant we wouldn't add this in the Lite version repo, but that we should actually write the code into the Pro version; which results in Lite+ being generated from it; i.e., build a Lite+ variation of this into the Pro copy.

Ugh, that's getting confusing. haha!

raamdev:
> build a Lite+ variation of this into the Pro copy.

Given that there _isn't_ a Lite+ version yet, I wasn't even considering that version. However, my feeling is that this is a simple feature that **does not** need to go into the Lite version, given how little of the work from this PR cab be reused due to the different codebase. (The original feature request, which was intended for Lite and Pro, was created _before_ the ZenCache codebase was forked.)

All of that said, when we do release a Lite+ version, I would love to see this feature available in Lite+.

What I can do is add an `if(IS_PRO){}` check so that we can get this out with the next Pro release, and then leave the original feature request open, with a new `lite+` tag, so that we remember to come back and add this feature at some point after releasing Lite+. How does that sound?

jaswsinc:
> What I can do is add an if(IS_PRO){} check so that we can get this out with the next Pro release, and then leave the original feature request open, with a new lite+ tag, so that we remember to come back and add this feature at some point after releasing Lite+. How does that sound?

Yep, that sounds great. We could try to do the Lite+ variation right now and be done with it, but there is still a bit of an unknown when it comes to how we will pull the latest Lite+ variation, so what you're saying makes perfect sense to me, yes.

raamdev:
@jaswsinc @renzms Merged! Thanks for your work on this, Renz!

renzms:
@jaswsinc @raamdev , Thanks for all your help, learned a lot from this.

So the Lite/Lite+ version will be attended to some time in the future?

raamdev:
> So the Lite/Lite+ version will be attended to some time in the future?

Correct. There's nothing to do for that right now, at least not until after we've launched Lite+.

PR #143, opened 2015-09-04
Feature/427

raamdev:
See websharks/zencache#427

_This is a work in progress...._


raamdev:
@jaswsinc I'm copying over your list of "Suggested Next Actions" (from [here](https://github.com/websharks/zencache/issues/427#issuecomment-121787639)) for reference:

---

## Suggested Next Actions

- [x] Create a new sub-folder in [this directory](https://github.com/websharks/zencache-pro/tree/000000-dev/src/includes/templates) called: `htaccess`

- [x] Create several files in that directory. One for each of the items that we want to add to the WordPress `.htaccess` file. We can start in this issue by adding one named: `templates/htaccess/cdn-filters.txt`. That file should contain:
  
  ```apache
  <IfModule headers_module>
    </FilesMatch>
  </IfModule>
  ```

- [ ] Add a new class member [here](https://github.com/websharks/zencache-pro/blob/150626/src/includes/closures/Plugin/InstallUtils.php#L194) called `$self->addWpHtaccess()` and that can be used to gather the templates that should be installed (based on plugin configuration) and pop those into the root `.htaccess` file for WordPress automatically.

- [ ] Add another class member: `$self->removeWpHtaccess()` to remove those entries.

- [ ] With those methods in place, make a call to `$self->addWpHtaccess()` here: https://github.com/websharks/zencache-pro/blob/150626/src/includes/closures/Plugin/InstallUtils.php#L17

- [ ] Make a call to `$self->addWpHtaccess()` here also: https://github.com/websharks/zencache-pro/blob/150626/src/includes/closures/Plugin/InstallUtils.php#L46

- [ ] Make a call to `$self->removeWpHtaccess()` here: https://github.com/websharks/zencache-pro/blob/150626/src/includes/closures/Plugin/InstallUtils.php#L65

- [ ] Make a call to `$self->removeWpHtaccess()` here also: https://github.com/websharks/zencache-pro/blob/150626/src/includes/closures/Plugin/InstallUtils.php#L87

---

Note that `addWpHtaccess()` should start by making a call to `removeWpHtaccess()` first. That way what it adds is from a fresh start; i.e., based on the current configuration of ZenCache.

---

raamdev:
@jaswsinc I could use some help (or pointers) fleshing out `$self->addWpHtaccess()` and `$self->removeWpHtaccess()`.

jaswsinc:
After [this line](https://github.com/websharks/zencache-pro/blob/150821/src/includes/closures/Shared/FsUtils.php#L111) add the following new utility.

```php
/*
 * Finds absolute server path to `/.htaccess` file.
 *
 * @since 15xxxx Adding `.htaccess` tweaks.
 *
 * @return string Absolute server path to `/.htaccess` file;
 *    else an empty string if unable to locate the file.
 */
$self->findHtaccessFile = function () use ($self) {
    if (!is_null($file = &$self->staticKey('findHtaccessFile'))) {
        return $file; // Already cached this.
    }
    $file = ''; // Initialize.
    if (is_file($abspath_htaccess = ABSPATH.'.htaccess')) {
        $file = $abspath_htaccess;
    } elseif (is_file($dirname_abspath_htaccess = dirname(ABSPATH).'/.htaccess')) {
        $file = $dirname_abspath_htaccess;
    }
    return $file;
};
```

---

After [this line](https://github.com/websharks/zencache-pro/blob/150821/src/includes/closures/Plugin/InstallUtils.php#L197) add the following new utilities.

```php
/*
 * Add template blocks to `/.htaccess` file.
 *
 * @since 15xxxx Adding `.htaccess` tweaks.
 *
 * @return boolean True if added successfully.
 */
$self->addWpHtaccess = function () use ($self) {
    if (!$self->options['enable']) {
        return false; // Nothing to do.
    }
    if (!$self->removeWpHtaccess()) {
        return false; // Unable to remove.
    }
    if (!($htaccess_file = $self->findHtaccessFile())) {
        if (!is_writable(ABSPATH) || file_put_contents($htaccess_file = ABSPATH.'.htaccess', '') === false) {
            return false; // Unable to find and/or create `.htaccess`.
        } // If it doesn't exist, we create the `.htaccess` file here.
    }
    if (!is_readable($htaccess_file)) {
        return false; // Not possible.
    }
    if (($htaccess_file_contents = file_get_contents($htaccess_file)) === false) {
        return false; // Failure; could not read file.
    }
    $template_blocks = '# BEGIN '.GLOBAL_NS."\n"; // Initialize.

    if (is_dir($templates_dir = dirname(dirname(__FILE__)).'/templates/htaccess')) {
        foreach (scandir($templates_dir) as $_template_file) {
            switch ($_template_file) {
                case 'cdn-filters.txt':
                    if ($self->options['cdn_enable']) {
                        $template_blocks .= trim(file_get_contents($templates_dir.'/'.$_template_file))."\n";
                    } // Only if CDN filters are enabled at this time.
                    break;
            }
        }
        unset($_template_file); // Housekeeping.
    }
    $template_blocks        = trim($template_blocks)."\n".'# END '.GLOBAL_NS;
    $htaccess_file_contents = $template_blocks."\n\n".$htaccess_file_contents;

    if (stripos($htaccess_file_contents, GLOBAL_NS) === false) {
        return false; // Failure; unexpected file contents.
    }
    if (defined('DISALLOW_FILE_MODS') && DISALLOW_FILE_MODS) {
        return false; // We may NOT edit any files.
    }
    if (!is_writable($htaccess_file)) {
        return false; // Not possible.
    }
    if (file_put_contents($htaccess_file, $htaccess_file_contents) === false) {
        return false; // Failure; could not write changes.
    }
    return true; // Added successfully.
};

/*
 * Remove template blocks from `/.htaccess` file.
 *
 * @since 15xxxx Adding `.htaccess` tweaks.
 *
 * @return boolean True if removed successfully.
 */
$self->removeWpHtaccess = function () use ($self) {
    if (!($htaccess_file = $self->findHtaccessFile())) {
        return true; // File does not exist.
    }
    if (!is_readable($htaccess_file)) {
        return false; // Not possible.
    }
    if (($htaccess_file_contents = file_get_contents($htaccess_file)) === false) {
        return false; // Failure; could not read file.
    }
    if (stripos($htaccess_file_contents, GLOBAL_NS) === false) {
        return true; // Template blocks are already gone.
    }
    $regex                  = '/#\s*BEGIN\s+'.preg_quote(GLOBAL_NS, '/').'\b.*?#\s*END\s+'.preg_quote(GLOBAL_NS, '/').'\s*/is';
    $htaccess_file_contents = preg_replace($regex, '', $htaccess_file_contents);

    if (stripos($htaccess_file_contents, GLOBAL_NS) !== false) {
        return false; // Failure; unexpected file contents.
    }
    if (defined('DISALLOW_FILE_MODS') && DISALLOW_FILE_MODS) {
        return false; // We may NOT edit any files.
    }
    if (!is_writable($htaccess_file)) {
        return false; // Not possible.
    }
    if (file_put_contents($htaccess_file, $htaccess_file_contents) === false) {
        return false; // Failure; could not write changes.
    }
    return true; // Removed successfully.
};
```

---

You should be able to start work on `templates/htaccess/cdn-filters.txt` then, and you might want to review the `$self->addWpHtaccess()` method to see how this works.

jaswsinc:
@raamdev I'm thrilled to see you working on this. I'm going to hold-off on the `feature/sonic-boom` work that I was planning until this is implemented, because it will make it much easier for us to add tweaks powered by `mod_rewrite` in Apache.

raamdev:
@jaswsinc I checked off all the items on [this list](https://github.com/websharks/zencache-pro/pull/143#issuecomment-137641802). However, I'm thinking that we also need to add a call to `addWpHtaccess()` somewhere that gets triggered when options are saved, correct?

jaswsinc:
> somewhere that gets triggered when options are saved, correct?

Yes. You can review this block of code and duplicate this for the htaccess addition. The block for the htaccess addition should appear right after this. See: https://github.com/websharks/zencache-pro/blob/000000-dev/src/includes/classes/Actions.php#L405-L407

Same thing here: https://github.com/websharks/zencache-pro/blob/000000-dev/src/includes/classes/Actions.php#L451-L453

---

We also need to remove it as well. So review this block and duplicate this for htaccess removal.

See: https://github.com/websharks/zencache-pro/blob/000000-dev/src/includes/classes/Actions.php#L413-L415

And also here: https://github.com/websharks/zencache-pro/blob/000000-dev/src/includes/classes/Actions.php#L459-L461

raamdev:
@jaswsinc Done. I also added the necessary Dashboard notifications when unable to add/remove from the `.htaccess` file (see https://github.com/websharks/zencache-pro/commit/6a0c97c1df62e23f9874e1ab7ce042c36d7f2ad5) however the message is rather unhelpful right now:

> Failed to update your `/.htaccess` file automatically.

It would be nice if we could show the contents of the `.htaccess` template that we were try to add/remove from the `.htaccess` file so that the site owner could manually update that file. However, that would require updating `addWpHtaccess()` to return information about which template(s) we were trying to write. 

The other option would be to simply create a KB Article that explains where to find the various `.htaccess` file templates (`src/includes/templates/htaccess/`) and how to insert the relevant ones into your `.htaccess` file. Then the error message could point to that. Thoughts?

jaswsinc:
> The other option would be to simply create a KB Article that explains where to find the various .htaccess file templates (src/includes/templates/htaccess/) and how to insert the relevant ones into your .htaccess file. Then the error message could point to that. Thoughts?

Yeah, that seems fine to me.

> Failed to update your /.htaccess file automatically.

The most common reason (99.9% of the time) this will happen, is if the `.htaccess` file is not writable. So I'd suggest mentioning that the likely reason is that the file is not writable, and that they should make the file writable to resolve this. Once that has been resolved, the issue with them trying to manually figure out which bits and pieces they need to include in the `.htaccess` file becomes less of a problem.

raamdev:
@jaswsinc I updated the error message the following:

> Failed to update your `/.htaccess` file automatically. Most likely a permissions error. Please make sure it has permissions `644` or higher (perhaps `666`). Once you've done this, please try again.

jaswsinc:
Cool!

raamdev:
@jaswsinc I'm revisiting this for the first time in a few weeks. It looks like this is about ready to be closed, do you agree? 

I saw your comment with NGINX-equivalent rules (https://github.com/websharks/zencache/issues/427#issuecomment-141892814); since we can't write to NGINX config files, I assume the goal would be to simply show a Dashboard notice (or a notice inside the Static CDN Filters options panel) that provides the recommended NGINX config, correct? Or maybe a Dashboard notice + a link to a KB Article that provides the recommended snippet would be better...

Which reminds me: I just reviewed this PR and realized that we're assuming that so long as we find an `.htaccess` file, we're running Apache, which may not be the case (someone could have migrated their site from Apache to a server running NGINX and carried over an old `.htaccess` file, in which case we would be updating the `.htaccess` file and assuming that all is well). We should probably be doing some checks here (along with appropriate Dashboard messaging) when running on NGINX, correct?

jaswsinc:
> I assume the goal would be to simply show a Dashboard notice (or a notice inside the Static CDN Filters options panel) that provides the recommended NGINX config, correct? Or maybe a Dashboard notice + a link to a KB Article that provides the recommended snippet would be better...

That sounds like a great idea to me also.

> We should probably be doing some checks here (along with appropriate Dashboard messaging) when running on NGINX, correct?

Agree. WordPress itself has globals `$is_apache` and `$is_nginx` that could help.
https://github.com/WordPress/WordPress/blob/2c23ca7411ccd37d35f56d5058b8172d4664e65b/wp-includes/vars.php#L100

raamdev:
> That sounds like a great idea to me also.

Which idea? :-) Linking to a KB Article that contains the recommended NGINX config, or showing the recommended NGINX config directly inside the Dashboard?

jaswsinc:
> Linking to a KB Article that contains the recommended NGINX config

I'd vote for this one, because that would allow us to update it remotely.

raamdev:
@jaswsinc Could you please review https://github.com/websharks/zencache-pro/commit/1dfc6c7656519794b9b12fd0d4e8e2e0f257aa8a and https://github.com/websharks/zencache-pro/commit/53eb98ff099f4f98c7998b229a9b75d42cec1456 when you get a chance?

Here are a few screenshots:





The notice will appear in both places, just in case the site owner misses the one in the Static CDN Filters option panel.

Also, I'm thinking it's best to keep this notice and the KB Article as generic as possible, i.e., to cover all the scenarios where ZenCache may write to the `.htaccess` file (I'm thinking about future work we do to tweak the `.htaccess` file). That way, we can use a single KB Article (https://github.com/websharks/zencache-kb/issues/91) to document all of the recommended NGINX server config snippets related to each feature in ZenCache that would normally write to the `.htaccess` file. That seems better than trying to create separate KB Articles (and separate Dashboard messages) for each scenario.

raamdev:
Oh, also, any feedback on the wording of the notice itself? I was struggling to keep it short and informative while also being newbie-friendly.

jaswsinc:
> Could you please review 1dfc6c7 and 53eb98f when you get a chance?

Those look perfect to me. Wonderful! I like how you added the `$is_apache` checks to prevent unnecessary notices related to `.htaccess`.

> Also, I'm thinking it's best to keep this notice and the KB Article as generic as possible, i.e., to cover all the scenarios where ZenCache may write to the .htaccess file (I'm thinking about future work we do to tweak the .htaccess file). That way, we can use a single KB Article (websharks/zencache-kb#91) to document all of the recommended NGINX server config snippets related to each feature in ZenCache that would normally write to the .htaccess file. That seems better than trying to create separate KB Articles (and separate Dashboard messages) for each scenario.

I agree. That's great! I added an [additional comment](https://github.com/websharks/zencache-kb/issues/91#issuecomment-148276587) over with a more elaborate Nginx example.

raamdev:
@jaswsinc Because of what I discovered [here](https://github.com/websharks/zencache/issues/427#issuecomment-121774596) (namely that you must ensure that `Access-Control-Allow-Origin` headers are being sent _before_ ZenCache starts rewriting your URLs, otherwise the CDN may cache the fact that `Access-Control-Allow-Origin` is not being sent, resulting in the CORS error), does it make sense to make the Static CDN Filters notice a bit more prominent? 

If a site owner is using NGINX and enables Static CDN Filters _before_ updating their NGINX configuration, they may still run into issues (because of how some CDNs seem to cache the initial headers sent by the server). Right now I have the Static CDN Filters NGINX notice inside the section that is greyed out until you enable Static CDN Filters:



Maybe instead that should go directly above the dropdown where you enable Static CDN Filters, along with an additional note that says something like "It is recommended that you update your NGINX server configuration _before_ enabling Static CDN Filters to prevent any issues with your CDN." 

Thoughts?

jaswsinc:
> "It is recommended that you update your NGINX server configuration before enabling Static CDN Filters to prevent any issues with your CDN.

Yes. That sounds like an excellent idea to me also. In addition, the KB article could mention what you experienced and perhaps make the same suggestion about doing Nginx _before_ you enable Static CDN Filters for the first time.

raamdev:
@jaswsinc Please review https://github.com/websharks/zencache-pro/commit/dc99f61ad54b6200189f44882f7d670ab74a6eef when you get a chance. I've improved the notice, added links to additional KB Articles, and created a filter that allows a site owner to disable the notice if it bugs them after they've updated their NGINX config. :-)




- http://zencache.com/r/kb-article-recommended-nginx-server-configuration/
- http://zencache.com/r/kb-article-what-are-cross-origin-request-blocked-cors-errors/
- http://zencache.com/r/kb-article-how-do-i-disable-the-nginx-htaccess-notice/

jaswsinc:
Nice! I reviewed the latest changes and each of your KBA drafts.

The only thing I noticed were calls to `apply_filters()` in the menu page and in the `Actions{}` class. Nothing wrong with those of course, but you might consider changing `apply_filters()` to `$this->plugin->applyWpFilters()`. That's an internal wrapper that we have so that it makes it easier for us to deal with branding changes and back compat. when the name of the plugin changes in the future.

See: https://github.com/websharks/zencache-pro/blob/151004/src/includes/closures/Shared/HookUtils.php#L236

So for instance, instead of:

```php
apply_filters(GLOBAL_NS.'_wp_htaccess_nginx_notice', TRUE)
```

Write it as:

```php
$this->plugin->applyWpFilters(GLOBAL_NS.'_wp_htaccess_nginx_notice', true)
```

raamdev:
> you might consider changing `apply_filters()` to `$this->plugin->applyWpFilters()`

Ah, good idea. Thanks! https://github.com/websharks/zencache-pro/commit/ee29e02da8d59e1de23061f75c27fd0ebfa6f9de fixes that. :-)

jaswsinc:
:+1:

raamdev:
@jaswsinc FYI https://github.com/websharks/zencache-pro/commit/647f81f58e5c8fb3f1dd4f26ac53442ca5283302

jaswsinc:
Ah, good catch!

raamdev:
@jaswsinc I noticed that if WordPress Address is `http://example.com/wordpress` but the Site Address is `http://example.com/` (i.e., WordPress is installed in `/wordpress/` subdirectory), ZenCache is only writing to `/wordpress/.htaccess` and not `/.htaccess`, even though `/.htaccess` exists. 

Is this going to be an issue? 

As an aside, I noticed that s2Member is doing the same thing. 

Also, the main WordPress rewrite rules were written by WordPress to `/.htaccess`, not `/wordpress/.htaccess`, hence the confusion here.

raamdev:
I see [here](https://github.com/websharks/zencache-pro/blob/feature/427/src/includes/closures/Shared/FsUtils.php#L120-L128) that we're using `ABSPATH.'.htaccess'` if found and only looking for and using `dirname(ABSPATH).'/.htaccess'` if `ABSPATH.'.htaccess'` is not found. I guess I just want to make sure that's not a problem. 

If `http://example.com/some/post/` is requested, will the web server still use the `.htaccess` rules in `/wordpress/` (despite that not being part of the URL path) because `http://example.com/some/post/` is essentially being rewritten and routed through `/wordpress/` in the background?

raamdev:
> WordPress's `index.php` and `.htaccess` files should be together in the directory indicated by the Site address (URL) setting on your General Options page.

via https://codex.wordpress.org/Using_Permalinks#Where.27s_my_.htaccess_file.3F

That seems to indicate that best practice here would be to follow suit and look for and use the `.htaccess` file in the directory indicated by the Site address (URL) setting.

raamdev:
Here's how WordPress finds `.htaccess`: 

```php
$home_path = get_home_path();
$htaccess_file = $home_path.'.htaccess';
```
https://github.com/WordPress/WordPress/blob/6b449abadc6751d4d3d1a805fe522dede8e7f690/wp-admin/includes/misc.php#L169-L200

jaswsinc:
> Here's how WordPress finds .htaccess via save_mod_rewrite_rules():

Cool. Nice find! I hadn't seen that portion of the codebase before. Looks good to me.

raamdev:
@jaswsinc How's this look? https://github.com/websharks/zencache-pro/commit/912d196ae112cc657f6a425834d93716c3487774

jaswsinc:
Looks great to me. I was just reviewing that function [here](https://github.com/WordPress/WordPress/blob/6b449abadc6751d4d3d1a805fe522dede8e7f690/wp-admin/includes/file.php#L81) and it looks sweet!

src/includes/closures/Plugin/InstallUtils.php
```diff
@@ -200,6 +204,108 @@
...
+            return false; // Unable to find and/or create `.htaccess`.
+        } // If it doesn't exist, we create the `.htaccess` file here.
+    }
+    if (!is_readable($htaccess_file)) {
+        return false; // Not possible.
+    }
+    if (($htaccess_file_contents = file_get_contents($htaccess_file)) === false) {
+        return false; // Failure; could not read file.
+    }
+    $template_blocks = '# BEGIN '.GLOBAL_NS."\n"; // Initialize.
```
raamdev: @jaswsinc I noticed that we're using `GLOBAL_NS` here, instead of `NAME`, which seems more appropriate. (`# BEGIN ZenCache` looks nicer than `# BEGIN zencache`.) Is there any reason we're not using `NAME` here?

jaswsinc: Nope. No reason. I agree that would be better.

raamdev:
@jaswsinc: 

> Fatal error: Call to undefined function WebSharks\ZenCache\Pro\get_home_path() in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/zencache-pro/src/includes/closures/Shared/FsUtils.php on line 122

Hmm, why wouldn't `get_home_path()` be available here? Is that because this is inside a Closure?

jaswsinc:
> Hmm, why wouldn't get_home_path() be available here?

It looks like that function is actually admin-only; i.e., you will need to explicitly include this file and/or limit the use of that function to the admin context only. Referencing [this source file](https://github.com/WordPress/WordPress/blob/6b449abadc6751d4d3d1a805fe522dede8e7f690/wp-admin/includes/file.php#L81) in WordPress core where that function is defined.

---

**Note:** Files in the `closures/Shared` directory may run in both phases of ZenCache. So if you write anything into `FsUtils.php` that might depend on WordPress being fully loaded, it would best to move that code into the `closures/Plugin` directory and perhaps create a new file or add it to an existing file.

raamdev:
@jaswsinc What if we just move `findHtaccessFile()` to `closures/Plugin/InstallUtils.php` (that's where it's being used by `addWpHtaccess()` and `removeWpHtaccess()` anyway) and then copy over the bit of code from `get_home_path()`:

```php
/*
 * Finds absolute server path to `/.htaccess` file.
 *
 * @since 15xxxx Adding `.htaccess` tweaks.
 *
 * @return string Absolute server path to `/.htaccess` file;
 *    else an empty string if unable to locate the file.
 */
$self->findHtaccessFile = function () use ($self) {
    $file = ''; // Initialize.

    $home    = set_url_scheme( get_option( 'home' ), 'http' );
    $siteurl = set_url_scheme( get_option( 'siteurl' ), 'http' );
    if ( ! empty( $home ) && 0 !== strcasecmp( $home, $siteurl ) ) {
        $wp_path_rel_to_home = str_ireplace( $home, '', $siteurl ); /* $siteurl - $home */
        $pos = strripos( str_replace( '\\', '/', $_SERVER['SCRIPT_FILENAME'] ), trailingslashit( $wp_path_rel_to_home ) );
        $home_path = substr( $_SERVER['SCRIPT_FILENAME'], 0, $pos );
        $home_path = trailingslashit( $home_path );
    } else {
        $home_path = ABSPATH;
    }
    $home_path = str_replace( '\\', '/', $home_path );

    if (is_file($htaccess_file = $home_path.'.htaccess')) {
        $file = $htaccess_file;
    }
    return $file;
};
```

↑ I tested that and it's working as expected.

raamdev:
Or we could strip out all the htaccess-related methods from `InstallUtils.php` and create a new `HtaccessUtils.php` file in `closures/Plugin/`.

jaswsinc:
> I tested that and it's working as expected.

Cool!

> InstallUtils.php and create a new HtaccessUtils.php file in closures/Plugin/

Either of those sound good to me. Whichever seems better to you.

raamdev:
@jaswsinc Please review these three commits when you get a chance: https://github.com/websharks/zencache-pro/pull/143#commits-pushed-fcc8389.

I just ran through several tests of this branch on two different test sites and the adding/removing from the `.htaccess` file is working well. I also tested failures to add/remove due to permissions errors and those notices show up correctly.

What I haven't tested are any of the things related to running on an NGINX server.

This branch could use another tester, but otherwise it looks ready to merge to me. :-) I'll hold off until I get the OK from you.

jaswsinc:
I'm testing this on Nginx now and I'll report back shortly.

Everything else looks great! Only suggestion is that perhaps you add the `$is_apache` check to `addWpHtaccess()` also.

raamdev:
> Only suggestion is that perhaps you add the $is_apache check to addWpHtaccess() also.

I only didn't do that because `addWpHtaccess()` calls `removeWpHtaccess()` at the top and `removeWpHtaccess()` checks if `$is_apache`. 

However, I realize now that it would be better practice to add it to `addWpHtaccess()` too in case `addWpHtaccess()` ever gets refactored in the future and calling `removeWpHtaccess()` ends up _not_ happening first. :-)

jaswsinc:
I just ran some tests in Nginx. Looks great! Here are a couple of observations.

- The "update your server configuration manually" link should probably open in a new window.

  

- Same thing with this link. I sort of expected this to hide the notice when I clicked it, but opening a new window with the KBA should be good too.

  

- Same thing in the links in this notice.

  

---

I tested the details provided in this KBA and everythings works as expected: https://github.com/websharks/zencache-kb/issues/91

raamdev:
> I sort of expected this to hide the notice when I clicked it, but opening a new window with the KBA should be good too.

Yeah, I debated whether or not I should even add that link there for that very reason. I wanted a way of linking to the KB Article that describes how to hide that message, but maybe it's enough to say "you can safely ignore this message" and leave it at that? If someone gets annoyed by the message they can write in or search the KB and make use of the KB Article that way. 

I thought about making it easier to hide the message, but I actually think it's good that it's hard (i.e., that it requires creating an MU-Plugin). That message is saying something quite important, and we wouldn't want a novice site owner to easily dismiss the message without updating their configuration and then write to us wondering why things don't seem to be working properly.

raamdev:
> The "update your server configuration manually" link should probably open in a new window.
> Same thing with this link.
> Same thing in the links in this notice.

Ah, good catch. I'll fix that!

> I just ran some tests in Nginx. Looks great!

Sweet! Thanks so much for testing that. :-)

raamdev:
> we wouldn't want a novice site owner to easily dismiss the message without updating their NGINX configuration

On the other hand, once the NGINX configuration _has_ been updated, it would be quite annoying (and misleading) to continue seeing that red error notice showing up in the ZenCache options...

jaswsinc:
Let me see if there is some way we can get a flag in our Nginx config. example that we could detect in PHP. That might make it easier to automatically deactivate that notice.

jaswsinc:
I updated the [example here](https://github.com/websharks/zencache-kb/issues/91#issue-111540221) to include the following. If they actually do add this in, you can detect it in PHP by looking at `$_SERVER['WP_NGINX_CONFIG'] === 'done'`

```nginx
  location ~* \.php$ {
      fastcgi_param WP_NGINX_CONFIG done;
  }
```

raamdev:
@jaswsinc Sweet! I updated the notices to check for `WP_NGINX_CONFIG`; see https://github.com/websharks/zencache-pro/commit/aa6674d47300b5c124aee40eb7a5433e6c4b46b0

If you could give that one more test on your NGINX site with `WP_NGINX_CONFIG` defined to confirm it's working as expected, that would be great.

jaswsinc:
Lookin' good on my side. Working as expected, yes.

raamdev:
@jaswsinc Cool! So this PR (_finally_) looks ready to merge. Agreed?

jaswsinc:
Yep. Looks ready to me. Really nice work on this!

raamdev:
@jaswsinc writes...
> I'm going to hold-off on the feature/sonic-boom work that I was planning until this is implemented, because it will make it much easier for us to add tweaks powered by mod_rewrite in Apache.

Just a reminder that this has now been implemented and merged. :-)

> Really nice work on this!

Thank you! This was fun to work on. :smile:

jaswsinc:
> Just a reminder that this has now been implemented and merged. :-)

Very cool. Thanks for the reminder.

PR #151, opened 2015-10-06
Admin Toolbar cached when Logged-In Ussers caching enabled

renzms:
Admin Toolbar cached when Logged-In Ussers caching enabled

; See: websharks/zencache#497


src/includes/classes/Plugin.php
```diff
@@ -490,6 +490,12 @@ public function setup()
         /*[/pro]*/
 
         /*[pro strip-from="lite"]*/
+        if ($this->options['when_logged_in'] === '1' && $this->applyWpFilters(GLOBAL_NS.'_when_logged_in_no_admin_bar', true)) {
+          show_admin_bar(FALSE); // Prevent admin bar from being cached.
```
raamdev: @jaswsinc It might be better to use `add_filter('show_admin_bar', '__return_false');` here instead of `show_admin_bar(FALSE)`, as the latter hides the Admin Bar from the front-end _and_ the backend (Dashboard). Since ZenCache is only affecting the front-end of the site, there's no need to also get rid of it on the Dashboard.

jaswsinc: @raamdev I think the docs on that `show_admin_bar()` function are a little outdated now. I just worked with trying to hide the admin bar whenever I built the demo.s2member.com site. I had to jump through some hoops in order to hide it in the Dashboard. It seems they have made it impossible to hide the admin bar in the Dashboard with this function or with the filter. The only way I was able to hide it in the Dashboard was by removing the rendering hook and by also injecting some custom CSS into the head.

Referencing: https://github.com/websharks/demos.websharks-inc.com/blob/master/wp-content/mu-plugins/demo.s2member.com.php#L39

Referencing: https://github.com/websharks/demos.websharks-inc.com/blob/master/wp-content/mu-plugins/demo.s2member.com.php#L161

raamdev: @jaswsinc Ah, good to know. Cool. In that case `show_admin_bar()` should be fine here. :-)

src/includes/classes/MenuPageOptions.php
```diff
@@ -522,8 +522,14 @@ public function __construct()
             echo '      <p><select name="'.esc_attr(GLOBAL_NS).'[saveOptions][when_logged_in]">'."\n";
             echo '            <option value="0"'.selected($this->plugin->options['when_logged_in'], '0', false).'>'.__('No, do NOT cache; or serve a cache file when a user is logged-in (safest option).', SLUG_TD).'</option>'."\n";
             echo '            <option value="postload"'.selected($this->plugin->options['when_logged_in'], 'postload', false).'>'.__('Yes, and maintain a separate cache for each user (recommended for membership sites).', SLUG_TD).'</option>'."\n";
-            echo '            <option value="1"'.selected($this->plugin->options['when_logged_in'], '1', false).'>'.__('Yes, but DON\'T maintain a separate cache for each user (I know what I\'m doing).', SLUG_TD).'</option>'."\n";
+            if ($this->plugin->options['when_logged_in'] === '1' || get_site_option(GLOBAL_NS.'_when_logged_in_was_1')) {
+                update_site_option(GLOBAL_NS.'_when_logged_in_was_1', '1');
```
raamdev: @jaswsinc This new `_when_logged_in_was_1` option is preparation for "Existing users who have this option enabled" as discussed in https://github.com/websharks/zencache/issues/497#issuecomment-125353814, correct?

jaswsinc: Correct. That flag should allow site owners who were using this feature to continue to have it as an available option in their Dashboard, even though it is being removed by default now.

raamdev:
@jaswsinc My review of this PR is finished and it looks ready to merge. Do you concur?

jaswsinc:
I just reviewed this also. Looks good to me, yes.

raamdev:
@renzms Merged! Thanks so much! :-)

@jaswsinc Thanks for the review and answering my questions. :-)

PR #155, opened 2015-10-08
Convert line-delimited patterns to a regex; See: websharks/zencache#191

kristineds:
Convert line-delimited patterns to a regex.
- Enhancing exclusion patterns

See: websharks/zencache#191

PR #159, opened 2015-10-16
Automatically clears cache files for a custom list of URLS; See: websharks/zencache#111

kristineds:
Automatically clears cache files for a custom list of URLS; 
-  Could specify a custom list of URIs that should be purged when a post is published/modified

See: websharks/zencache#111

### DIFF
--- a/src/includes/classes/MenuPageOptions.php
+++ b/src/includes/classes/MenuPageOptions.php
@@ -375,6 +375,11 @@ class MenuPageOptions extends MenuPage
             echo '      <p class="info" style="display:block; margin-top:-15px;">'.__('In a Multisite Network, each child blog (whether it be a sub-domain, a sub-directory, or a mapped domain); will automatically change the leading <code>http://[sub.]domain/[sub-directory]</code> used in pattern matching. In short, there is no need to add sub-domains or sub-directories for each child blog in these patterns. Please include only the <a href="https://gist.github.com/jaswsinc/338b6eb03a36c048c26f" target="_blank">REQUEST_URI</a> (i.e., the path) which leads to the XML Sitemap on all child blogs in the network.', SLUG_TD).'</p>'."\n";
         }
         echo '      </div>'."\n";
+        if (IS_PRO || $this->plugin->isProPreview()) {
+            echo '<h4 style="margin-bottom:0;">'.__('Auto-Clear a List of Custom URLs Too?', SLUG_TD).'</h4>'."\n";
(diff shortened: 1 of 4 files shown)


kristineds:
@jaswsinc @raamdev Submitting pull request for feature request:  User-defined URIs to purge with post cache

jaswsinc:
@kristineds Nice work. This looks good to me.

raamdev:
@kristineds Thank you!

I noticed the new **Auto-Clear a List of Custom URLs Too?** option appears to be listed under the **Sitemap-Related Options** section, which is not correct. 



There should be a new `<hr>` and new `<h3>` heading (like Sitemap-Related Options) called **Misc. Auto-Clear Options** that makes it clear the new option is not part of the Sitemap-Related Options:


 
---

@jaswsinc By looking at the code it appears that wildcards and regex patterns are _not_ supported in this list of Custom URLs to auto-clear, is that correct? If so, @kristineds we should add a note after "Please list one URL per line." that clarifies this: "(Note that wildcards and RegEx's are **not** supported here.)"

raamdev:
Also noting here that supporting wildcards in this Custom URLs to auto-clear box would make this feature solve various other scenarios and really increase the power of this feature. See https://github.com/websharks/zencache/issues/111#issuecomment-133766571

Thoughts @jaswsinc?

jaswsinc:
> supporting wildcards in this Custom URLs

Agreed. My suggestion would be to exclude that from this PR (i.e., after Kristine makes the changes you suggested above, go ahead and merge it). Then we can make the regex enhancement a part of the work being done in this other PR instead. See: https://github.com/websharks/zencache-pro/pull/155#issuecomment-148803566

That way it will support the new improved flavor as well.

raamdev:
> My suggestion would be to exclude that from this PR (i.e., after Kristine makes the changes you suggested above, go ahead and merge it). Then we can make the regex enhancement a part of the work being done in this other PR instead. See: #155 (comment)

Sounds great! Thanks! :smiley:

kristineds:
@raamdev @jaswsinc Re-submitting PR which includes the following: https://github.com/websharks/zencache-pro/commit/746b27b1cd0faee3830238c35726ae7f4aa382b2
- Add new heading Misc. Auto-Clear Options 
- Add text: `(Note that wildcards and RegEx's are **not** supported here.)`

raamdev:
@kristineds Perfect! Thanks so much. :-) Merged!

PR #166, opened 2015-10-23
Strip whitespace from contents of wp-config.php when checking WP_CACHE

raamdev:
See websharks/zencache#591

### DIFF
--- a/src/includes/closures/Plugin/InstallUtils.php
+++ b/src/includes/closures/Plugin/InstallUtils.php
@@ -134,7 +134,7 @@ $self->addWpCacheToWpConfig = function () use ($self) {
     if (!is_readable($wp_config_file)) {
         return ''; // Not possible.
     }
-    if (!($wp_config_file_contents = file_get_contents($wp_config_file))) {
+    if (!($wp_config_file_contents = php_strip_whitespace($wp_config_file))) {
         return ''; // Failure; could not read file.
     }
     if (preg_match('/define\s*\(\s*([\'"])WP_CACHE\\1\s*,\s*(?:\-?[1-9][0-9\.]*|TRUE|([\'"])(?:[^0\'"]|[^\'"]{2,})\\2)\s*\)\s*;/i', $wp_config_file_contents)) {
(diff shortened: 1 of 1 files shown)


raamdev:
@jaswsinc Review please. I tested this and it's working as expected when enabling/disabling ZenCache.

raamdev:
Thinking this over some more, it's probably less confusing to only modify the `preg_match()` line and pass in `php_strip_whitespace($wp_config_file)` instead of `$wp_config_file_contents`.

That way we can leave `if (!($wp_config_file_contents = file_get_contents($wp_config_file))) {` alone. I was modifying that line because I was thinking of the performance hit reading the file twice, but that's not really an issue here (as this is only run when enabling/disabling ZenCache).

jaswsinc:
Nice! Yeah, I agree with your thoughts on there being a separate call to read the file like that.

raamdev:
> I agree with your thoughts on there being a separate call to read the file like that.

So you think it should remain as-is, or I should change things by moving the `php_strip_whitespace()` call to `preg_match()`?

jaswsinc:
> So you think it should remain as-is, or I should change things by moving the php_strip_whitespace() call to preg_match()?

I would create a second variable with a duplicate block like this one. See: https://github.com/websharks/zencache-pro/blob/be0a44af033adc05e3f3074e58ac1381e76d29ee/src/includes/closures/Plugin/InstallUtils.php#L137-L139

One block for `$wp_config_file_contents` and another for `$wp_config_file_contents_no_whitespace`. That way you are checking if `$wp_config_file_contents_no_whitespace` is empty also, and you can use `$wp_config_file_contents_no_whitespace` in your `preg_match()` test.

raamdev:
> You could do this on both of these lines also. See: https://github.com/websharks/zencache-pro/blob/be0a44af033adc05e3f3074e58ac1381e76d29ee/src/includes/closures/Plugin/InstallUtils.php#L185-L188

Can't do it there because that would end up writing the whitespace-stripped version of `wp-config.php` back to the file...

I did make a few improvements as you suggested though; see https://github.com/websharks/zencache-pro/commit/639a7ad7cc7e7a9b7e8c1cb795293bb9d3453db4. How's that look?

jaswsinc:
> see 639a7ad. How's that look?

Looks great to me! :-)

> You could do this on both of these lines also. See: https://github.com/websharks/zencache-pro/blob/be0a44af033adc05e3f3074e58ac1381e76d29ee/src/includes/closures/Plugin/InstallUtils.php#L185-L188
> Can't use it in the preg_replace() call because that would end up writing the whitespace-stripped version of wp-config.php back to the file...

I was referring the `\b` in front `define` there.

raamdev:
> I was referring the \b in front define there.

Oh, whoops, didn't notice that was a commit comment. :-) Done: https://github.com/websharks/zencache-pro/commit/cea249bce6dd9ce77db538bf0757f240238d5bce

jaswsinc:
Awesome. Looks ready to merge to me.

raamdev:
Thanks! :-)

PR #169, opened 2015-10-30
PR: feature/596

jaswsinc:
See: websharks/zencache#596

## TODO

- [x] See: https://github.com/websharks/zencache-pro/pull/169#issuecomment-152667334
- [x] Add screenshot previews of each so that you get a visual whenever you change the dropdown value.
- [ ] Add a new clearing option (or config. option) that allows for clearing by Post ID.


jaswsinc:
---

jaswsinc:
@raamdev This should be ready to merge. Please review when you have some time.

jaswsinc:
## Things to know about the split button options list.

- The "Current Page Only" option only shows on the front-end of the site. You won't see this while working in the Dashboard; i.e., in `/wp-admin/*` areas.

- The "OPCache Only" option only shows if OPCache is running on the server, and only if the current user can `update_plugins` or `manage_network_plugins` on a MS network.

- The "CDN Only" option only shows if Static CDN Filters are enabled, and only if the current user can `update_plugins` or `manage_network_plugins`.

- The "Specific Page Only" option requires that you enter a specific URL. Upon clicking this option you will be prompted to enter the URL. If you don't, nothing happens.

jaswsinc:
Noting that this PR does not cover _all_ of the ideas presented in https://github.com/websharks/zencache/issues/596#issuecomment-151321067. However, it does implement some of the most-requested items. We can add new options in the future.

clavaque:
Very nice! :D

raamdev:
@jaswsinc Woooohoooo! Nice! 

Here's some initial feedback after my testing of this branch.

### New Menu

The `border-left` of the dropdown arrow visually separates it from the "Cache Stats" and "Clear Cache" buttons, which alienates it. Furthermore, a border on the left without a border on the right, makes it appear to be part of the "Howdy, raamdev" section. I recommend changing it to a `border-right`.



That said, my preference would be to get rid of that dropdown arrow altogether. Enabling the dropdown options should simply cause the menu of options to appear when you hover over the existing "Clear Cache" button. 

Clicking the Clear Cache button would still clear the entire cache, but instead you could just hover over it and then click one of the other options underneath. That seems to be a far more intuitive way to implement those additional options, because when you separate them from the existing "Clear Cache" button, it sort of feels like you need to reiterate what the options in that new menu actually do (you tried doing this by repeating "Clear cache:" inside the menu, but that seems inadequate and feels odd.

Consolidating that menu also gets rid of the additional clutter being added to the Admin Bar (we already have two buttons taking up space there) and creates the same effect as the Cache Stats button (i.e., hover over it to see something--it's a familiar behavior). 



Assuming we consolidate the menu, I would then suggest we change the menu item names to the following:

- Home Page Cache
- Current Page Cache
- Specific URL Cache
- OPCache
- CDN Cache

There's no need to specify "only" as describing the item being cleared implies that "only" the cache for that item is being cleared. 

### Clear Current Page vs Clear Current Post ID

> The "Current Page Only" option is limited to only the current page.

My feeling is that instead of a whole other option in the menu, we simply have an option inside Manual Cache Clearing that allows you to control the behavior of the "Clear Current Page Cache" menu option. You could set if it should only clear the current URL, or if it should clear everything related to that post (associated tag/category archives, etc.).

Either that, or we just change the behavior of the "Clear Current Page Cache" option so that it _always_ clears everything related to the current Page (assuming the current page is a Post/Page, otherwise it clears it like clearing a specific URL). If someone only wants to clear the specific URL for a Post/Page (i.e., they don't want everything related to the Post/Page to be cleared as well), then they should just use the "Clear Specific URL Cache" feature and specify the Post/Page URL. 

### Other notes

- In the JS popup that accepts a URL, instead of "Enter a specific URL" it should say **Enter a specific URL to clear the cache for that page**.
- If you enter a dummy URL (e.g., `http://blahblahblah123.com`) or even an invalid URL (e.g., `asdasdasd`) in the Specific URL JS popup, you still get a message that says "Cache successfully cleared", which seems wrong. Instead, it should say something like "You can only clear URLs on this site".

jaswsinc:
> Clicking the Clear Cache button would still clear the entire cache, but instead you could just hover over it and then click one of the other options underneath. That seems to be a far more intuitive way

I disagree on it being more intuitive. I did try this way first though, because I do agree that it is _cleaner_. However, in terms of it being more intuitive, I found that by nesting the additional options into the "Clear Cache" button itself, it produces the following issues:

- It's no longer obvious that you can simply click the "Clear Cache" button itself to clear the cache, because anytime you hover over the button you get a dropdown. If you're new to ZenCache and it's your first time seeing this button, I feel there are high odds that you'll think the only available options are those that you see in the dropdown menu; and you won't necessarily think to click the button itself.

- For those who do figure that out, or have used ZenCache in the past (like us), I find it slightly annoying that I have to see that dropdown selection of options each time I want to simply clear the cache; i.e., if I try to click the "Clear Cache" button itself, I still get the dropdown because I need to hover it in order to click it.

---

> a border on the left without a border on the right, makes it appear to be part of the "Howdy, raamdev" section. I recommend changing it to a border-right.

Great points. I will give this a shot first, as this is also a very easy change.

---

> There's no need to specify "only" as describing the item being cleared implies that "only" the cache for that item is being cleared.

I see what you mean. I agree with you.

> In the JS popup that accepts a URL, instead of "Enter a specific URL" it should say Enter a specific URL to clear the cache for that page.

Agreed.

---

> If you enter a dummy URL (e.g., http://blahblahblah123.com) or even an invalid URL (e.g., asdasdasd) in the Specific URL JS popup, you still get a message that says "Cache successfully cleared", which seems wrong. Instead, it should say something like "You can only clear URLs on this site".

I agree with this too. The problems with validating the URL, are:

- On a multisite network, you may have multiple sub-domains or even have domain mapping enabled. So for us to accurately validate the URL that is entered would require quite a bit of overhead in code and additional complexity that is not necessarily warranted in this case.

- On a standard install, there is also the possibility of a mapped domain. It's not a common, but there are sites running a standard WordPress installation with multiple DB table prefixes and/or host names (e.g., `sub.domain.com`, or `www.domain.com` vs. without, or `en.domain.com`, etc). So validating the URL that is given could potentially limit what sites like that could do in terms of clearing the cache.

_Perhaps what we can do is detect that it's a standard WP installation, and if the domain in the URL is not a match to the domain in `home_url()`, then we can add an additional note underneath the response that ZenCache reports back. That way at least ZenCache will be smart enough to acknowledge that the URL you cleared the cache for was not a match to the primary domain for the site. Thoughts on this? Any suggestions for what that warning/message should say exactly?_

jaswsinc:
Better 1?



---

Better 2?



---

Better 3?

raamdev:
> I feel there are high odds that you'll think the only available options are those that you see in the dropdown menu; and you won't necessarily think to click the button itself.

Then just add another option to the menu: "Clear Entire Cache"

> I find it slightly annoying that I have to see that dropdown selection of options each time I want to simply clear the cache;

Then disable the additional Clear Cache options inside Manual Cache Clearing... if you don't want to see those options, they can be hidden.

---

 > Better 1?

This looks like the current style, so it looks like the arrow is part of the "Howdy, jaswsinc".

 > Better 2?

This looks the best of the three, but I still think the arrow needs to go altogether. 

 > Better 3?

This makes the arrow look like its for an entirely different plugin, and then it looks wrong because it has no description.

---

I feel pretty strongly about that arrow + the Clear Cache button. It just seems so confusing to me, and I'm someone who uses ZenCache all the time!

raamdev:
> Perhaps what we can do is detect that it's a standard WP installation, and if the domain in the URL is not a match to the domain in home_url(), then we can add an additional note underneath the response that ZenCache reports back. That way at least ZenCache will be smart enough to acknowledge that the URL you cleared the cache for was not a match to the primary domain for the site. Thoughts on this? Any suggestions for what that warning/message should say exactly?

Yeah, that sounds like a good alternative to the current behavior. When the domain in the URL is not a match to the domain in `home_url()` on a standard WP installation, I would add the following note to the bottom of the notification:

> **Notice:** The domain you entered did not match your configured WordPress Home URL.

jaswsinc:
Minor tweaks review...

jaswsinc:
> Then just add another option to the menu: "Clear Entire Cache"

That could work to resolve this issue, I agree.

> Then disable the additional Clear Cache options inside Manual Cache Clearing... if you don't want to see those options, they can be hidden.

That doesn't work for me, because perhaps I occasionally want to clear the OPCache Only, or the CDN Cache Only, and I really appreciate having those in the admin bar and readily available. However, I only do that occasionally, so I don't want to see those options all the time; i.e., I also appreciate the simplicity of the "Clear Cache" button as a stand-alone feature, which is a sentiment that you shared a while back whenever we discussed this before. Giving site owners some control over how this button is introduced and in how the additional options are presented could be a small, but significant peeve for some people.

> This makes the arrow look like its for an entirely different plugin, and then it looks wrong because it has no description.

Hmm. Well I can agree with that, I don't consider that as important myself, because once you hover over this for the first time and see that it's for clearing the cache with additional options, then there is no confusion any longer.

> I feel pretty strongly about that arrow + the Clear Cache button. It just seems so confusing to me, and I'm someone who uses ZenCache all the time!

Same here. I feel the opposite of you in this regard. However, that difference between us tells me this is going to be a personal preference. Having it in the "Clear Cache" button bugs me, not having it in there bugs you. I wonder if there is some way for us to make that configurable also.

raamdev:
> I wonder if there is some way for us to make that configurable also.

Yep, that's exactly what I was thinking would be a simple solution to this. It could simply be an extra option in Manual Cache Clearing:

jaswsinc:
Great. Yep, that should do it. Adding a TODO entry above.

jaswsinc:
:laughing:

raamdev:
:smile_cat: :rage: :see_no_evil:

jaswsinc:
@raamdev Please review when you get a moment. This should be ready to merge now.

> - [ ] Add a new clearing option (or config. option) that allows for clearing by Post ID.

I'm punting this item in my TODO list, for now. I think it needs some more thought myself, but if you disagree then let me know what you'd prefer and I'll see about it. I read your thoughts on this above, but my general feeling is we need user feedback on that part.

I actually think that the current behavior is what will be expected, and that clearing anything more than a specific page is not desirable. Also, if you change a Post/Page the cached is cleared automatically already, so I don't see a lot to be gained from us trying to clear by post ID in this list of options. In this particular context, my feeling is that a single item being cleared is much easier to comprehend and to build trust in as you learn to use these cache clearing options over time.

raamdev:
> I actually think that the current behavior is what will be expected, and that clearing anything more than a specific URL is not desirable. Also, if you change a Post/Page the cache is cleared automatically already, so I don't see a lot to be gained from us trying to clear by post ID in this list of options. In this particular context, my feeling is that a single item being cleared is much easier to comprehend and to build trust in as you learn to use these cache clearing options over time.

I agree. I'd say lets leave the behavior as it is now (clearing a single URL and nothing more).

I'll review this again shortly. Thank you for splitting the menu options. :-) I think that was a win for users as it lets people pick what they want and doesn't really add much overhead or confusion to anything.

raamdev:
@jaswsinc I just spent some more time testing this. It looks and works great! Thank you.

There was one minor annoyance that I corrected in https://github.com/websharks/zencache-pro/commit/4f5a79937dbd331b151ba21e5d11c4db4e975699, along with a few other slight changes:

- I removed the `title` value from the Clear Cache button when in a standard WP installation. I found that the title there can get annoying and in the way, especially when your intention is to view or choose an item from the menu that comes below the the Clear Cache button. (The response to seeing the menu appear when hovering over "Clear Cache" is to stop moving the mouse to look at the menu options, but then the Clear Cache title appears, and distracts you from looking at the menu options.) My other justification for removing the `title` value is that it adds nothing that isn't already clearly apparent. The button already says "Clear Cache", so repeating "Clear Cache (Start Fresh)" in the title attribute seems rather redundant. In a Multisite installation there _is_ additional information that is helpful to include in a title, so I left that alone. (You'll notice that all of the other WordPress items in the Admin Bar follow the same logic: There are no title attributes for items that are clearly explained by their label, but for things like the Plugin Updates and the Comments icon, which _are not_ very descriptive, there is a title that explains how many plugin updates or comments there are.)
- I removed the `title` attribute from the Cache Stats button for the reason described above.
- I _added_ several `title` attributes to the new clear cache options in the new menu (as those would benefit from a bit of explanation):
     - **Home Page**: "Clear the Home Page cache"
     - **Current URL**: "Clear the cache for the current URL"
     - **Specific URL**: "Clear the cache for a specific URL"
     - **OPCache**: "Clear PHP's OPCache"
     - **CDN Cache**: "Clear the CDN cache"
- I left alone the `title` attribute on the new split options menu arrow icon, as that clearly needs something descriptive to indicate it's a menu of "Clear options".

jaswsinc:
Cool. Nice work on this enhancements!

raamdev:
> Things to know about the split button options list.

I wrote a KB Article to document some of those (it will be marked as published after the release; see todo [here](https://github.com/websharks/zencache.com/issues/73#issuecomment-152791488)):

- [Why are Clear Cache menu options missing?](https://github.com/websharks/zencache-kb/issues/96)

I also added a todo item to this Watered-down Syntax KB Article to mention support for that syntax in the "Specific URL" option: https://github.com/websharks/zencache-kb/issues/94#issuecomment-152791544

PR #170, opened 2015-11-01
Update build script to use new phings submodule

raamdev:
See websharks/phings#15 websharks/zencache#579 websharks/zencache.com#72


raamdev:
@jaswsinc: In https://github.com/websharks/phings/pull/15#issuecomment-143332784 you said: 
> Add a new submodule to bring the websharks/phings project in.
> ```
$ git submodule add git@github.com:websharks/phings.git phings
> ```

But won't that just be getting the `000000-dev` branch of the `phings` repo? I see you have a Phings v151002 release here: https://github.com/websharks/phings/releases 

However, by simply adding the phings repo as a submodule to the ZenCache Pro repo, aren't we just using the dev branch of the Phings repo? It seems we should make sure we're _not_ doing that, as when we start working on improving the build scripts to do other fancy stuff in the future, we wouldn't want to inadvertently start using that code in to build projects _until_ we deem it ready and release it.

jaswsinc:
What you can do is this...

```
$ git submodule add git@github.com:websharks/phings.git phings
$ cd phings
$ git checkout master
```

Actually, it's also worth noting that Git Submodules are never associated with any specific branch or tag. They are simply a reference to a particular snapshot (i.e., an  `@sha`). They remain that way unless and until you cd into the submodule directory and change the pointer; e.g., changing the branch or running  `git pull` to update them in some way.

So running...

```
$ git submodule add git@github.com:websharks/phings.git phings
```

That will give you a Git Submodule that references the latest dev branch at the time that you created the submodule, and then it will stay that way even if the remote repo changes, and even if you run `git submodule update` from ZenCache Pro in the future. The only way the `@sha` pointer changes is if you navigate to the submodule directory and move the sha by changing branches or pulling to update.

You will know when the submodule sha changes, because when/if that changes, running a `git status` from the ZenCache Pro repo will show that your submodule pointer has moved, and that's a commit that has to be recorded against the ZenCache Pro repo itself.

raamdev:
@jaswsinc Excellent explanation! Thank you. :smile:

PR #176, opened 2015-11-16
Check if running latest RC; See: websharks/zencache#352

kristineds:
Check if running latest RC; 

- This is Step 1 of 2. Note: Don't merge yet until step 2 has been completed
- Add option `pro_update_check_stable`

See: websharks/zencache#352


raamdev:
@jaswsinc @kristineds As I [noted in the other PR](https://github.com/websharks/zencache-pro/pull/177#issuecomment-158823200), I've merged the commits from the Step 2 PR (https://github.com/websharks/zencache-pro/pull/177) into this Step 1 PR and will be testing and leaving further updates here. If there are changes to commit regarding this feature, please commit them to the `feature/352-step1` branch.

Also, I merged in the latest `000000-dev` branch, so when you get a chance please update your local copy of `feature/352-step1` (you can run `git fetch origin && git checkout feature/352-step1 && git pull`).

raamdev:
@jaswsinc I built and tested this branch (which now includes changes from https://github.com/websharks/zencache-pro/pull/177 as noted above), however when I enable Beta Updates and then click "Update Now", ZenCache is still pulling the latest public version (v151114) instead of the RC version that I uploaded (filename `zencache-pro-v151129.rc.zip`; note that the filename is v151129 but the actual plugin version in the code inside that zip file is v151123; I changed the filename to something a few days ahead for testing purposes). Despite that newer RC filename existing on ZenCache.com, the version installed by the Pro updater is always v151114.

jaswsinc:
> instead of the RC version that I uploaded (filename zencache-pro-v151129.rc.zip

I see. A while back, you and I discussed what suffix you could give RC copies of the software, so that you could _absolutely_ avoid them ever showing up in any automated scan that might release them. I suggested that you use `.rc`, because according to semantic versioning rules, the `.rc` development suffix is invalid. Thus, if our automated routines are following the proper guidelines, a release file with that slightly invalid name would not show up. I think this is why you're having trouble.

Now that we have decided to _include_ RC releases through the API (not just as direct downloads), you will need to name the RC files according to semantic versioning guidelines. See: http://semver.org/

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92.

So use: `zencache-pro-v151129-rc.zip` (then wait 10 minutes before trying again).

_**Note:** Our system will also allow for no dash, as that is a WP/PHP standard that we bend for. So while no dash is not technically valid, we allow this to come through also: `zencache-pro-v151129rc.zip`. I prefer the dash be there, but it's good to note this minor deviation from the official guidelines._

raamdev:
> So use: zencache-pro-v151129-rc.zip (then wait 10 minutes before trying again).

But doesn't that mean that everyone who currently has the Pro Updater active and configured will suddenly be prompted to update? (i.e., regular users who _don't_ want to install an RC, who are currently running ZenCache Pro v151114 or earlier, will the Pro Updater as released with v151114, see the `zencache-pro-v151129-rc.zip` file and think that it's a new official release?)

raamdev:
Actually, re-reading what you wrote [here](https://github.com/websharks/zencache/issues/352#issuecomment-140816558), I see that the Product API is what controls what is "seen" by the Pro Updater in the plugin. So I assume this means that older versions of the software (that don't report `stable=1`) will still get the latest stable version because the Product API defaults to `stable=1`.

raamdev:
@jaswsinc I just tried renaming the file to `zencache-pro-v151129-rc.zip` and waiting 15-20 minutes: When I run the Pro Updater with "Yes, check for release candidates", it still downloads and installs `zencache-pro-v151114.zip`.

jaswsinc:
> I see that the Product API is what controls what is "seen" by the Pro Updater in the plugin.

Correct. That was updated to help separate stable releases from development-state suffixes like `-rc`, `-alpha`, `-beta`. So the default configuration in ZenCache Pro (the software) is to set `stable=1`, and that is the API default at ZenCache.com as well. So the only way to actually get development copies is to set `stable=0` explicitly, which is what this PR is making it possible for site owners to do.

jaswsinc:
> I just tried renaming the file to zencache-pro-v151129-rc.zip and waiting 15-20 minutes: When I run the Pro Updater with "Yes, check for release candidates", it still downloads and installs zencache-pro-v151114.zip.

I can confirm that the API itself is returning the RC release now as expected. Could you try it once more and see if it is still broken? If so, the problem must be in the work we are doing here; i.e., in the ZenCache Pro software itself, because the API is picking this up correctly.

raamdev:
> Could you try it once more and see if it is still broken?

Just tried again--it still updates to v151114.

Note that when I "enable" Release Candidate checks, I'm simply visiting the Pro Updater, changing the dropdown in the new Beta Updates panel to "yes, check for release candidates" and then hitting "Update Now" (there is no way to "save" my change). Maybe this isn't working because the update is being run with the default configuration (i.e., only stable updates)?

jaswsinc:
Thank you trying that again. I will run some tests to see if you're right.

jaswsinc:
You were right. Bug fixed.

raamdev:
> You were right. Bug fixed.

Yep, that was it. I just retested and it's working great now. :-) Thanks!

raamdev:
@jaswsinc I'm ready to merge this into `000000-dev` for further testing, unless you have anything you'd like to add to this PR first...

jaswsinc:
Nope. Looks good to me.

raamdev:
Merged! Wooohooo! Awesome work @kristineds! :+1:

kristineds:
:+1:

PR #178, opened 2015-11-17
Bug Fix: Installing, activating, deactivating, and deleting a plugin does not wipe/clear the cache; see: websharks/zencache#424

renzms:
see: websharks/zencache#424


raamdev:
@jaswsinc Can we add a filter here to disable this behavior should a site owner want to do so?

raamdev:
@jaswsinc Ping. ↑

jaswsinc:
@raamdev I see no problem with that.

## @renzms Next Actions

- Before [this line](https://github.com/websharks/zencache-pro/blob/6eeeb6f2d50240351ba5496a08929ef315d9f16e/src/includes/closures/Plugin/WcpPluginUtils.php#L18) add the following:

  ```php
  if (!$self->applyWpFilters(GLOBAL_NS.'_auto_clear_on_plugin_activation_deactivation', true)) {
      return 0; // Nothing to do here.
  }
  ```

renzms:
@jaswsinc /@raamdev

PR submitted!

raamdev:
@renzms Thanks! Tested and merged. :-)

jaswsinc:
:+1:

PR #179, opened 2015-11-18
Bug Fix - Pagination Var Translation Breaks Cache Clearing

renzms:
see:  websharks/zencache#607

### DIFF
--- a/src/includes/closures/Shared/CachePathUtils.php
+++ b/src/includes/closures/Shared/CachePathUtils.php
@@ -2,6 +2,39 @@
 namespace WebSharks\ZenCache\Pro;
 
 /*
+* Cache-path suffix frag (regex).
+*
+* @since 15xxxx Enhancing translation support.
+*
+* @param string $regex_suffix_frag Existing regex suffix frag?
+*
+* @return string Cache-path suffix frag (regex).
+*/
+$self->cachePathRegexSuffixFrag = function ($regex_suffix_frag = CACHE_PATH_REGEX_DEFAULT_SUFFIX_FRAG) use ($self) {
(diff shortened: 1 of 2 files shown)


src/includes/closures/Shared/CachePathUtils.php
```diff
@@ -2,6 +2,39 @@
...
+
+/*
+* Default cache-path suffix frag (regex).
+*
+* @since 15xxxx Enhancing translation support.
+*
+* @return string Default cache-path suffix frag (regex).
+*/
+$self->cachePathRegexDefaultSuffixFrag = function () use ($self) {
+    if ($self->isPlugin() && !empty($GLOBALS['wp_rewrite'])){
```
raamdev: @jaswsinc I see here that we're only checking if `!empty($GLOBALS['wp_rewrite'])` and then assuming `$GLOBALS['wp_rewrite']->pagination_base` and `$GLOBALS['wp_rewrite']->comments_pagination_base` have valid values. My intuition tells me that might not be the case (i.e., what if `pagination_base` contained a valid value, but `comments_pagination_base` was empty?). 

I haven't reviewed enough of the WP Rewrite code to know for sure, but I wanted to double-check with you before merging this PR.

jaswsinc: I think it's fine. `$GLOBALS['wp_rewrite']` refers to an instance of the `WP_Rewrite` class, and those pagination values are both properties of that class; i.e., expected to always exist.

raamdev: > I think it's fine. $GLOBALS['wp_rewrite'] refers to an instance of the WP_Rewrite class, and those pagination values are both properties of that class; i.e., expected to always exist.

Copy that. Thank you.

PR #180, opened 2015-11-18
Add UI option for cache cleanup schedule; See: websharks/zencache#472

kristineds:
Add UI option for cache cleanup schedule; 

- This is Step 2 of 2 of ZC GHI: `purge_cache` CRON Should Run Every Hour 
- Added under Cache Expiration Time per Raam's instructions. **ZenCache → Plugin Options → Cache Expiration Time**
- Removed any mention of the Auto-Cache Engine

ZenCache Cache Cleanup Schedule

See: websharks/zencache#472


kristineds:
@jaswsinc @raamdev Please see latest commit with updated changes https://github.com/websharks/zencache-pro/commit/feeebd98e0b28e211f9c6cd9dd047d065e9b8857

raamdev:
@kristineds Nice work! Here's a small requested change:

The description currently reads: "If you have an extremely large site with a lower expiration time, %1$s can be told to set a Cache Cleanup Schedule to prevent the cache files from slowing down your server via <a href="https://codex.wordpress.org/Category:WP-Cron_Functions" target="_blank" style="text-decoration:none;"><em>WP Cron</em></a> at a <code>hourly</code> <em>(set by default)</em> interval."

I suggest we tweak that as follows:

"If you have an extremely large site and you set a lower cache expiration time, expired cache files can build up more quickly. By default, %1$s cleans up expired cache files via <a href="https://codex.wordpress.org/Category:WP-Cron_Functions" target="_blank" style="text-decoration:none;">WP Cron</a> at an <code>hourly</code> interval, but you can tell %1$s to use a custom Cache Cleanup Schedule below to run the clean up process more or less frequently depending on your specific needs."

kristineds:
@raamdev That has been updated. https://github.com/websharks/zencache-pro/commit/9d9195967afb8ee85cdce648a75775b0a320eac7

Placing it above the section titled _Disable Cache Expiration If Server Load Average is High?_

ZenCache Cache Cleanup Schedule

raamdev:
@kristineds Can you please remove the italic `<em>` around "WP Cron"? Also, if that link as `text-decoration: none`, please remove that as well.

Also, please change "and you set a lower cache expiration time" to "and you lower the default Cache Expiration Time of `7 days`".

kristineds:
@raamdev Latest commit submitted ;) https://github.com/websharks/zencache-pro/commit/d702eeb6858b8e209e94be68497c268fa4609711

- removed the italic <em> around "WP Cron"
- removed as text-decoration: none,
- changed "and you set a lower cache expiration time" to "and you lower the default Cache Expiration Time of 7 days".
- the following text in Cache Expiration Time updated to reflect the new hourly schedule (it currently says "once daily", it should now say "once per hour"):

ZenCache Cache Cleanup

jaswsinc:
Some minor tweaks from me in https://github.com/websharks/zencache-pro/commit/0d2357bde304ee385a69860da186625738b57460

- Changed `clean up` to `cleanup`.
- Made "WP Cron" link a Redirect from zencache.com.
- Made `wp_schedule_event()` link a Redirect from zencache.com.
- Changed "compatible with WordPress `wp_schedule_event()`" to improve wording just a bit.

jaswsinc:
@raamdev Would it be a good idea to make this a select menu of predefined options? That would eliminate the need for the additional **Tip** below this field, and we could pull the list of available options using: https://codex.wordpress.org/Function_Reference/wp_get_schedules

My feeling is that having a select menu would make this more user-friendly. That way, what you select is going to work, no question. The default CRON schedules are: `hourly`, `twicedaily` and `daily`. With ZenCache installed, you pick up one more: `every15m`; i.e., it is added by ZenCache, and it will show up in a call to `wp_get_schedules()`.

raamdev:
@jaswsinc writes...
> Would it be a good idea to make this a select menu of predefined options? That would eliminate the need for the additional Tip below this field, and we could pull the list of available options using: https://codex.wordpress.org/Function_Reference/wp_get_schedules

Yes! Great idea. That would be _much_ better here, especially given the fact that some site owners may have used another plugin to create their own WP Cron schedules and if we build a dynamic list using `wp_get_schedules()`, our list will also show their custom schedules.

raamdev:
@kristineds writes...
> Latest commit submitted. cc @jaswsinc d702eeb

Looks great. Thank you! :smile: 

@jaswsinc writes...
> Some minor tweaks from me in 0d2357b

Awesome. Thanks for those.

jaswsinc:
@raamdev Please review when you get a moment. Thanks :-)

raamdev:
@jaswsinc Sweet! I just tested the menu and it looks great! One question: The default when I installed this branch version was "Twice Daily"; is that correct? I thought it was being changed to "hourly".

jaswsinc:
> The default when I installed this branch version was "Twice Daily"; is that correct? I thought it was being changed to "hourly".

Cool. Yeah, I had some unexpected results like that also. If you can merge this into the dev branch, along with step 1 also, we can test both of those together. I think testing this branch by itself as-is, is problematic, because it doesn't include the new option key internally, which goes with this UI enhancement.

raamdev:
> I think testing this branch by itself as-is, is problematic, because it doesn't include the new option key internally, which goes with this UI enhancement.

Ah, that makes sense. I had assumed this branch did include those changes. :-) Thanks. I'll merge these in shortly.

raamdev:
Merged. Thank you @kristineds and @jaswsinc! :-)

kristineds:
Cool @jaswsinc! Love those dropdown menu! Definitely more user-friendly than typing it in :+1:

jaswsinc:
> Love those dropdown menu! Definitely more user-friendly than typing it in

Cool. Glad to hear it! :-)

PR #182, opened 2015-11-18
Feature Request: URI Exclusion Patterns for Client-Side Cache;

renzms:
see websharks/zencache#498


jaswsinc:
This looks great to me! Nice work @renzms :100:

raamdev:
@renzms Sweet! I'm looking forward to seeing a PR for Step 2. :-)

raamdev:
@jaswsinc As per my feedback to Renz (see https://github.com/websharks/zencache/issues/498#issuecomment-159713315), I suggested that we not make this an enable/disable option. Instead, it should work the same way URI Exclusion Patterns work, where the exclusion patterns will simply be used if there are any included in the box (assuming that Client-Side Caching is also enabled). I think this PR will need to be updated to include those changes (i.e., get rid of the enable/disable option for this). Can you let Renz know what changes will be necessary in this PR?

jaswsinc:
@raamdev Copy that. Thank you.

I see that you wrote previously... @raamdev writes...

> Remove the option to enable/disable this feature--if there are exclusion patterns in the box, they will be used, otherwise not (the same way URI Exclusion Patterns works).

---

### A quick review of what we have right now:

- Client-Side URI Exclusions are a separate field that you configure.
  See: **ZenCache → Plugin Options → Client-Side Cache**
- If entered, and if client-side caching is allowed, they will always be applied. Nothing extra to enable.

---

In addition to this, we have regular URI Exclusions that apply to server-side caching.
  See: **ZenCache → Plugin Options → URI Exclusions**

- If you enter anything there, and you have client-side caching enabled, ZenCache can be told to _also_ apply your regular URI Exclusions too; i.e., make them a part of your Client-Side URI Exclusions also, without you needing to enter them all over again.

  So _this_ is where we have an enable/disable option.

  That enable/disable option was based on the comment that @raamdev made here.

  > https://github.com/websharks/zencache/issues/498#issue-87504848
  > It would be nice if there were a few extra options inside the existing URI Exclusion Patterns panel (such as an option to "Also exclude these URLs from Client-Side Caching"), as well as an additional section with a box for listing URI Exclusion Patterns that should apply only to Client-Side Caching.

---

### My Question

- Are you saying that we should just _assume_ that any of the regular URI Exclusion patterns should be applied to client-side exclusions also? i.e., without the enable/disable option?

raamdev:
> Are you saying that we should just assume that any of the regular URI Exclusion patterns should be applied to client-side exclusions also? i.e., without the enable/disable option?

@jaswsinc @renzms Sorry for all the confusion here. :-) My mind wasn't fully made up about how this should work or look when I opened the GitHub issue so now that I'm seeing this implemented here's what I feel we should do:

### Client-Side Caching Options Panel

- Add "Client-Side Caching Exclusion Patterns" box that allows a site owner to list exclusions. This box should always be visible, but the exclusion patterns will only be used if Client-Side Caching is enabled.

### URI Exclusion Patterns Options Panel

- DON'T add any new options to the URI Exclusions Patterns options panel.
- DO make URI Exclusion Patterns apply to Client-Side Caching. The current description in the URI Exclusion Patterns box says (emphasis mine) "where a particular file, or a particular group of files, **should never be cached.**" That to me would indicate that the URI Exclusion Patterns also apply to Client-Side Caching, i.e., if I have Client-Side Caching enabled and I exclude `/example/` in URI Exclusion Patterns, I would expect that Client-Side Caching will also be disabled for `/example/` ("should never be cached", client-side caching being a form a caching).
- Add a filter that allows an advanced site owner to disable this default behavior (i.e., make it possible for them to _not_ have URI Exclusion Patterns apply to Client-Side Caching). 

---

I thought about adding a dropdown to the URI Exclusion Patterns box (and was changing my mind even as I typed the above list) that would allow a site owner to include/exclude URI Exclusion Patterns from Client-Side Caching, but this is somewhere that I really feel it's important for us to make ["decisions, not options"](https://wordpress.org/about/philosophy/). Adding an option to the URI Exclusion Patterns panel adds confusion to the mix. A site owner who knows they just want to disable caching for `/example/` shouldn't learn about another type of caching in that panel (Client-Side Caching) and then wonder, "Hmm, do I want to leave this enabled or disabled??" Their goal is to disable caching for a particular page.

jaswsinc:
@raamdev Thanks! Crystal clear now, and I completely agree with that approach also. Let's leave this open and I'll come back and outline the remaining work here for @renzms soon.

jaswsinc:
## @renzms Additional Changes Needed in PR for Step 1

- Start from your existing work in `feature/498-step1`

- Remove [this line](https://github.com/websharks/zencache-pro/blob/5ef2732c0ecfe4b9721aae4fab0eb07686831f2e/src/includes/classes/Plugin.php#L314)

- Change [this line](https://github.com/websharks/zencache-pro/blob/5ef2732c0ecfe4b9721aae4fab0eb07686831f2e/src/includes/closures/Plugin/InstallUtils.php#L295) to the following:

  ```php
  if($self->applyWpFilters(GLOBAL_NS.'_ exclude_uris_client_side_too', true)) {
  ```

- Commit and push those changes to this same branch; i.e., `feature/498-step1`.

- Ask us for another please. Thanks for your patience here! :-)

renzms:
@jaswsinc 
PR submitted, please review. Thanks!

jaswsinc:
@renzms Looks great!

raamdev:
@renzms Please `git pull` from inside the `feature/498-step1` branch and then make all future commits to that branch. I have merged `feature/498-step2` into this branch, so once you update your local copy, you shouldn't need to use the Step 2 branch anymore. Committing any future work for this issue to one branch will help me with testing (so I don't need to do a manual merge before testing each time).

---

@renzms When I tested this, entering anything (e.g., `/category/*`) into the new Exclusion Patterns for Client-Side Caching box and then clicked "Save All Changes" at the bottom, the changes did not save, i.e., the box was empty the next time I looked at it.

It seems there is a bug somewhere. 

I haven't looked for it yet, but I thought I'd let you try bug hunting a bit first to see if you can find the bug. :-)

renzms:
@raamdev 

Figured out what was wrong, in MenuPageOptions,

I had this line `$this->plugin->options['exclude_uris']` instead of 

`$this->plugin->options['exclude_client_side_uris']`. So during testing, I must've saved something in URI Exclusion Patterns and it saved successfully as it was the same value! :)

Thanks for catching that one. Should be working now!

plugin options renz s test area wordpress 4

Submitting PR.

raamdev:
@renzms Merged. Awesome work. Thank you. :smile:

PR #185, opened 2015-11-21
Enhance logged-in user caching support; See: websharks/zencache#601

kristineds:
Enhance logged-in user caching support; 

- Add `nc_debug_is_logged_in_user_nonce` and `nc_debug_page_contains_nonce`
- This is Step 1 of 2. _Note: Don't merge yet until Step 2 has been completed._

See: websharks/zencache#601


kristineds:
@jaswsinc @raamdev Submitted PR for step 1 of 2. https://github.com/websharks/zencache-pro/commit/8c105241c1b6b54f9f753b65a9368326e3840edb

Any update on how I should be tackling step 2 of 2 for this? :)

jaswsinc:
@kristineds I don't think there will need to be a step 2 after all. I added some additional tweaks in this PR just a moment ago, and that should do it here, I think. Pending a review from @raamdev.

@raamdev This should be ready to merge now. We now disable the Akistment nonce check whenever caching is enabled, as described [here](https://github.com/websharks/zencache/issues/601#issuecomment-156429335) in our previous discussion.

## PR Summary

- ZenCache will no longer cache pages that contain `_wpnonce` in the markup, no matter if logged-in user caching is enabled or not.

- ZenCache will no longer cache pages that contain `akismet_comment_nonce` in the markup, no matter if logged-in user caching is enabled or not.

- ZenCache will automatically disable `akismet_comment_nonce` as discussed [here](https://github.com/websharks/zencache/issues/601#issuecomment-156429335), which will avoid a problematic scenario with Akismet; i.e., the second item listed here is a moot point now.

  This new default behavior can be controlled via a new filter:

  _Default behavior is true; i.e., we do disable the Akistment nonce. Set false to change it._

  ```php
  add_filter('zencache_disable_akismet_comment_nonce', '__return_false');
  ```

kristineds:
@jaswsinc:
> I don't think there will need to be a step 2 after all. I added some additional tweaks in this PR just a moment ago, and that should do it here, I think

Got it. Thanks for the update. ;)

raamdev:
@kristineds Merged. Thank you! :smile:

PR #187, opened 2015-11-24
Add unique key to BEGIN/END comment blocks in .htaccess

raamdev:
See websharks/zencache#620

### DIFF
--- a/src/includes/classes/VsUpgrades.php
+++ b/src/includes/classes/VsUpgrades.php
@@ -46,6 +46,7 @@ class VsUpgrades extends AbsBase
         $this->fromQuickCache();
         $this->fromLte150807();
         $this->fromLte151107();
+        $this->fromLte151114();
     }
 
     /**
@@ -309,6 +310,59 @@ class VsUpgrades extends AbsBase
                     $this->plugin->updateOptions($this->plugin->options); // Save/update options.
                     $this->plugin->activate(); // Reactivate plugin w/ new options.
                 }
+            }
+        }
+    }
+
+    /**
+     * Before we changed the htaccess comment blocks to contain a unique identifier.
+     *
+     * @since 15xxxx Improving `.htaccess` tweaks.
+     */
+    protected function fromLte151114()
+    {
+        if (version_compare($this->prev_version, '151114', '<=')) {
+            global $is_apache;
+
+            if (!$is_apache) {
+                return false; // Not running the Apache web server.
+            }
+            if (!($htaccess_file = $this->plugin->findHtaccessFile())) {
+                return true; // File does not exist.
+            }
+            if (!is_readable($htaccess_file)) {
+                return false; // Not possible.
+            }
+            if (($htaccess_file_contents = file_get_contents($htaccess_file)) === false) {
+                return false; // Failure; could not read file.
+            }
+            if (stripos($htaccess_file_contents, 'ZenCache') === false) {
+                return true; // Template blocks are already gone.
+            }
+            $v151114_htaccess_template_blocks = '# BEGIN ZenCache
+<IfModule headers_module>
+  <FilesMatch "\.(ttf|ttc|otf|eot|woff|woff2|font.css|css|js)$">
+    Header set Access-Control-Allow-Origin "*"
+  </FilesMatch>
+</IfModule>
+# END ZenCache
+';
+            $v151114_htaccess_template_blocks_disabled = '# BEGIN ZenCache
+# END ZenCache
+';
+            $regex                  = '/'.preg_quote($v151114_htaccess_template_blocks, '/').'/';
+            $htaccess_file_contents = preg_replace($regex, '', $htaccess_file_contents);
+            $regex                  = '/'.preg_quote($v151114_htaccess_template_blocks_disabled, '/').'/';
+            $htaccess_file_contents = preg_replace($regex, '', $htaccess_file_contents);
+
+            if (defined('DISALLOW_FILE_MODS') && DISALLOW_FILE_MODS) {
+                return false; // We may NOT edit any files.
+            }
+            if (!is_writable($htaccess_file)) {
+                return false; // Not possible.
+            }
+            if (file_put_contents($htaccess_file, $htaccess_file_contents) === false) {
+                return false; // Failure; could not write changes.
             }
         }
     }
--- a/src/includes/classes/VsUpgrades.php
+++ b/src/includes/classes/VsUpgrades.php
@@ -325,44 +325,32 @@ class VsUpgrades extends AbsBase
             global $is_apache;
 
             if (!$is_apache) {
-                return false; // Not running the Apache web server.
+                return; // Not running the Apache web server.
             }
             if (!($htaccess_file = $this->plugin->findHtaccessFile())) {
-                return true; // File does not exist.
+                return; // File does not exist.
             }
             if (!is_readable($htaccess_file)) {
-                return false; // Not possible.
+                return; // Not possible.
             }
             if (($htaccess_file_contents = file_get_contents($htaccess_file)) === false) {
-                return false; // Failure; could not read file.
+                return; // Failure; could not read file.
             }
             if (stripos($htaccess_file_contents, 'ZenCache') === false) {
-                return true; // Template blocks are already gone.
+                return; // Template blocks are already gone.
             }
-            $v151114_htaccess_template_blocks = '# BEGIN ZenCache
-<IfModule headers_module>
-  <FilesMatch "\.(ttf|ttc|otf|eot|woff|woff2|font.css|css|js)$">
-    Header set Access-Control-Allow-Origin "*"
-  </FilesMatch>
-</IfModule>
-# END ZenCache
-';
-            $v151114_htaccess_template_blocks_disabled = '# BEGIN ZenCache
-# END ZenCache
-';
-            $regex                  = '/'.preg_quote($v151114_htaccess_template_blocks, '/').'/';
-            $htaccess_file_contents = preg_replace($regex, '', $htaccess_file_contents);
-            $regex                  = '/'.preg_quote($v151114_htaccess_template_blocks_disabled, '/').'/';
-            $htaccess_file_contents = preg_replace($regex, '', $htaccess_file_contents);
-
+            if (is_dir($templates_dir = dirname(dirname(__FILE__)).'/templates/htaccess/back-compat')) {
+                $htaccess_file_contents = str_replace(file_get_contents($templates_dir.'/v151114.txt'), '', $htaccess_file_contents);
+                $htaccess_file_contents = str_replace(file_get_contents($templates_dir.'/v151114-2.txt'), '', $htaccess_file_contents);
+            }
             if (defined('DISALLOW_FILE_MODS') && DISALLOW_FILE_MODS) {
-                return false; // We may NOT edit any files.
+                return; // We may NOT edit any files.
             }
             if (!is_writable($htaccess_file)) {
-                return false; // Not possible.
+                return; // Not possible.
             }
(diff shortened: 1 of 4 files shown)


raamdev:
@jaswsinc A review here, please. Here's the regex on regex101.com that I'm using in this PR: https://regex101.com/r/vZ3rN0/1

jaswsinc:
Awesome! I just reviewed the code and regex. Here's are my thoughts.

- The end of your regex has `.*\s*`, and that `.*` is a problem. It will eat up anything after and until the end of the entire file, because of the `/s` flag; i.e., `.` dot matches newlines too, and it is greedy.

- I think the regex could be simpler if the marker is used exclusively as the start and end points, instead of it being a part of something more in terms of how of we search/replace.

  See possible alternative here: https://regex101.com/r/vZ3rN0/2
  
  Summary of Changes:

  - Strip the `=`, which is not required to properly decode base64.
  - Use `WmVuQ2FjaGU` as the open tag (like HTML), and `/WmVuQ2FjaGU` as the close tag (like HTML).
  - Simplify regex: `#\s*WmVuQ2FjaGU.*?\/WmVuQ2FjaGU\s*/s`
  - Drop the `/i` flag, because base64 is case sensitive.

raamdev:
> Use WmVuQ2FjaGU as the open tag (like HTML), and /WmVuQ2FjaGU as the close tag (like HTML).

Hmm, it would look a lot more readable if the comment started with `# BEGIN ZenCache`. Also, I was trying to make those comments _a little_ flexible, in case a site owner wanted to clean up his/her `.htaccess` file and maybe just use ` # BEGIN ZenCache WmVuQ2FjaGU`. I agree with all the other points--thank you!

jaswsinc:
> Hmm, it would look a lot more readable if the comment started with # BEGIN ZenCache. Also, I was trying to make those comments a little flexible

I see. Yeah, I agree. Well, since `#` is also a required part of the marker, we could just alter it slightly so that the open tag is actually `#.*?WmVuQ2FjaGU`. That way the open tag could be:

```
#[ANYTHING]WmVuQ2FjaGU
```

jaswsinc:
I think what is most important is that `/WmVuQ2FjaGU` be the very end.

raamdev:
> I think what is most important is that /WmVuQ2FjaGU be the very end.

What's the importance of `/`? Why couldn't we use `# END[anything]WmVuQ2FjaGU[whitespace]`?

jaswsinc:
Thats works too, so not important if you use `END` as the end marker.

My thought was that, since we are now using a unique marker instead of a specific phrase, the `/` could be used instead of a word like `END`. Thus, `/WmVuQ2FjaGU` would suffice, where `/` is used together with the marker so that a site owner could alter the words `BEGIN` and `END` without damaging anything.

raamdev:
Right, but `BEGIN` and `END` are far more common as markers, both in HTML comments and in `.htaccess` files (other plugins that write to the `.htaccess` file also use `BEGIN/END`.

jaswsinc:
You could still type those in there for clarity so that they appear in the `.htaccess` file, but because we are using a unique marker, which other plugins that _only_ use `BEGIN` and `END` do not have, we can get away with something a bit more elegant. In our case, the marker is not `BEGIN` `END`, the _real_ marker is `WmVuQ2FjaGU`, which uniquely identifies the automated portion of the `.htaccess`, to help separate it from anything that a site owner adds on their own.

So my feeling is that if `WmVuQ2FjaGU` is the _real_ marker, that, and only that (if possible), is what our regex should be looking for IMO; i.e., `WmVuQ2FjaGU` (open) and `/WmVuQ2FjaGU` (close), as an example. Or you could think about it like this, and maybe this is even better for clarity.

```
# <WmVuQ2FjaGU>
  Whatever here...
# </WmVuQ2FjaGU>
```

raamdev:
I hear your point, and from a programatic perspective that makes more sense. But from a readability perspective (which is what comments are designed for), it makes more sense to clearly indicate "this is where something begins and this is where something ends". Using an HTML style format inside a comment inside an `.htaccess` file is _confusing_ to say the least. Using a `/WmVuQ2FjaGU` is also confusing because then there's a question about whether the unique marker is `WmVuQ2FjaGU` or `/WmVuQ2FjaGU`.

All of that is resolved by following the informal style of `# BEGIN` and `# END`, including the name of the plugin (`ZenCache`) and including the unique marker `WmVuQ2FjaGU`. I would also include in the comment a note explaining that `WmVuQ2FjaGU` _is_ a unique marker required for the plugin to identify that codeblock (as in my original commit), because again, comments are designed to be human readable and informative. I don't think it should be _required_ to have that descriptive text in the codeblock, which is why I suggest we only require `#` `BEGIN/END`, `ZenCache`, and `WmVuQ2FjaGU` on the comment lines.

jaswsinc:
> All of that is resolved by following the informal style of # BEGIN and # END, including the name of the plugin (ZenCache) and including the unique marker WmVuQ2FjaGU. I would also include in the comment a note explaining that WmVuQ2FjaGU is a unique marker required for the plugin to identify that codeblock (as in my original commit), because again, comments are designed to be human readable and informative. I don't think it should be required to have that descriptive text in the codeblock, which is why I suggest we only require # BEGIN/END, ZenCache, and WmVuQ2FjaGU on the comment lines.

No objection. Sounds fine to me :-) I've heard it both ways. haha

raamdev:
@jaswsinc How's this? https://regex101.com/r/vZ3rN0/4

We can require the following format as a minimum:

```
# BEGIN ZenCache WmVuQ2FjaGU

# END ZenCache WmVuQ2FjaGU
```

An additional comment after the unique string explaining not to remove the unique string can be added by the plugin, i.e., 

```
# BEGIN ZenCache WmVuQ2FjaGU (do not remove WmVuQ2FjaGU marker)
```

raamdev:
Also, for usability reasons, I would make that regex case `i`nsensitive. Accepting `# begin zencache wmvuq2fjagu` should be OK too, IMO. But I'm fine with keeping it case-sensitive if you prefer that.

jaswsinc:
Looks great to me!

> Also, for usability reasons, I would make that regex case insensitive. Accepting # begin zencache wmvuq2fjagu should be OK too, IMO. But I'm fine with keeping it case-sensitive if you prefer that.

Agree, caSe-insensitive does sound better.

I wonder if it would help others to recognize `WmVuQ2FjaGU` as being a marker if it was encapsulated in some way. At the moment, it's just hanging out all by itself. Do you think wrapping it with something like `<WmVuQ2FjaGU>` or `{WmVuQ2FjaGU}` would make any difference?

raamdev:
My feeling is that encapsulating it distorts _what_ the unique key actually is (does it include `[` and `]`?), whereas surrounding it by space characters sort of clarifies that.

I would use the following as the first comment:

```
# BEGIN ZenCache WmVuQ2FjaGU (the WmVuQ2FjaGU marker is required for ZenCache; do not remove)
```

jaswsinc:
Copy. Looks nice.

raamdev:
@jaswsinc ↑ Review please.

jaswsinc:
Cool. I just took another look at this and I have one more suggestion and a question.

- I suggest generating `WmVuQ2FjaGU` based on `NAME`, so that whenever the software is rebranded that marker will change. Does sound sound right? e.g., `$marker = trim(base64_encode(NAME), '=')`

- What happens whenever you already have the old snippet (without the marker), and then you upgrade from one version of the software to the next? As it exists at the moment, it looks like the old snippet (without the marker) would _not_ be removed, and yet the new one will be added on top of the old one.

raamdev:
> I suggest generating WmVuQ2FjaGU based on NAME, so that whenever the software is rebranded that marker will change. Does that sound right? e.g., $marker = trim(base64_encode(NAME), '=')

Agreed.



> What happens whenever you already have the old snippet (without the marker), and then you upgrade from one version of the software to the next? As it exists at the moment, it looks like the old snippet (without the marker) would not be removed; i.e., error during reactivation.

Thanks for reminding me about that (I had thought of that issue while working on this but then forgot about it). Yeah, the way I see it we have two options:

- Strip out any existing `# BEGIN ZenCache` block, thereby risking us deleting custom code where someone else used the same comment marker (this is the current bug we're trying to fix).
- OR, leave any existing `# BEGIN ZenCache` block, thereby risking us duplicating the code we're inserting (their `.htaccess` could end up with two blocks of code that were added by ZenCache).

My vote is that we look for a _very specific_ block of code to narrow down what ZenCache v151114 may have added, i.e.,

```
# BEGIN ZenCache

<IfModule headers_module>
  </FilesMatch>
</IfModule>

# END ZenCache
```

If we find that specific block of code, with that specific `.htaccess` rule, _then_ we remove it and insert our block that includes the unique key.

raamdev:
We should also look for and remove this from the `.htaccess` file, which is what ZenCache v151114 would've added if you enabled Static CDN Filters and then disabled them:

```
# BEGIN ZenCache
# END ZenCache
```

jaswsinc:
> My vote is that we look for a very specific block of code to narrow down what ZenCache v151114 may have added, i.e.,

Ah, cool. Yep, that sounds like a good plan to me also.

raamdev:
> I suggest generating WmVuQ2FjaGU based on NAME, so that whenever the software is rebranded that marker will change. Does that sound right? e.g., $marker = trim(base64_encode(NAME), '=')

Actually, I was just thinking about this and maybe this _doesn't_ make sense. Why would we want a different marker if the name of the plugin changes? That will just mean we'll need to write upgrade routines that look for and remove the old blocks and then re-add them using the new unique marker. That seems like we're adding unnecessary code if we'll need to strip out those old blocks anyway to change the name of the plugin in those comment blocks.

If we leave the marker the same, then the only thing we need to do when changing the name of the plugin is to add 3 extra lines to `removeWpHtaccess()` that also looks for and removes the blocks that contain the old plugin name. It makes no difference if the unique marker is the same.

jaswsinc:
I see your point. Hard to predict exactly what would work best whenever we start on the rebranding, but I inclined to agree with you. Sounds like having just one would be better.

raamdev:
@jaswsinc  Another review please. ↑ I added an upgrade routine that will remove the old comment blocks as discussed above and I tested this code to confirm it works as expected when upgrading with either template scenario in place (empty or with the `Access-Control-Allow-Origin` code inside).

jaswsinc:
I haven't tested this yet, but I did review the code. Looks great!

Couple of observations...

- In `fromLte151114()`, the `return true;` and `return false;` lines could just be `return;`, because there is no need for the version-specific upgrade routines to return a boolean value, or return anything at all for that matter. A `return;` is the same as `return null;`, and that's what is expected from those routines.

- I wouldn't change anything else since you've already put in the work to test all of this. However, in the future you might consider putting verbatim data that you're going to search for in a text file, to keep it separate. It's less prone to corruption that way, and it keeps the codebase cleaner, since you won't need to single quote and wrap lines against the leftmost margin to get what you need.

  Also, you could just use `str_replace()` instead of `preg_replace()` whenver you're doing string-for-string search/replace. There is nothing wrong with using `preg_replace()`, but it's not really necessary in this instance either.

  ```php
  $htaccess_contents = str_replace(file_get_contents('find-this.txt'), '', $htaccess_contents);
  ```

raamdev:
Great feedback. Thank you. 

> you might consider putting verbatim data that you're going to search for in a text file

Where would you put such a text file? `src/includes/templates/htaccess/v151114.txt` and `v151114-2.txt`?

jaswsinc:
Yep, that looks good. Or, maybe inside a sub-directory that marks it as something old.

```
src/includes/templates/htaccess/back-compat/v151114.txt
```

raamdev:
@jaswsinc One more review, please. ↑ I tested the refactored code and it's working great.

jaswsinc:
Niiiice! :-) :heart_eyes: Love it.

PR #188, opened 2015-11-25
Only show welcome message when plugin is not already enabled

raamdev:
See websharks/zencache#622

### DIFF
--- a/src/includes/closures/Plugin/InstallUtils.php
+++ b/src/includes/closures/Plugin/InstallUtils.php
@@ -10,14 +10,17 @@ namespace WebSharks\ZenCache\Pro;
  */
 $self->activate = function () use ($self) {
     $self->setup(); // Ensure setup is complete.
-    if (!$self->options['welcomed']) {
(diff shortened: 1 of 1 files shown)


jaswsinc:
Ah, great catch. I think it should be this though. After the most recent change, new users won't see the message either, I think.

```php
if (!$self->options['enable'] && !$self->options['welcomed']) {
    $settings_url = add_query_arg(urlencode_deep(array('page' => GLOBAL_NS)), network_admin_url('/admin.php'));
    $self->enqueueMainNotice(sprintf(__('<strong>%1$s</strong> successfully installed! :-) <strong>Please <a href="%2$s">enable caching and review options</a>.</strong>', SLUG_TD), esc_html(NAME), esc_attr($settings_url), array('push_to_top' => true)));
    $self->updateOptions(array('welcomed' => '1'));
}
if (!$self->options['enable']) {
    return; // Nothing to do.
}
```

raamdev:
@jaswsinc Right you are. My subconscious was _just_ thinking about that. Thank you! Fixed in https://github.com/websharks/zencache-pro/commit/640781f04b9a54eae11287211468ba5038aff0d8

PR #189, opened 2015-11-26
Add UI option when Static CDN Filters are enabled for logged-in users; See: websharks/zencache#486

kristineds:
Add UI option when Static CDN Filters are enabled for logged-in users; 

-  Add new section to **ZenCache → Plugin Options → Logged-In Users** titled _Static CDN Filters Enabled for Logged-In Users & Comment Authors?_
- Add dropdown box with two options: 
  - No, disable Static CDN Filters when a user is logged-in (default option).
  - Yes, allow Static CDN Filters to be applied to logged-in users.

UI Option for when Static CDN Filters are enabled for logged-in users

See: websharks/zencache#486


kristineds:
@jaswsinc @raamdev Submitting PR for adding UI option on section:  _Static CDN Filters Enabled for Logged-In Users & Comment Authors?._ I would probably need some help with the text description. ;)

raamdev:
@kristineds Nice! :-) The wording is almost perfect. Here's what I recommend changing:

- [ ] I would remove this sentence and everything after it (it seems redundant to me): "This section describes how Static CDN Filters..."
- [ ] Change "When a user is logged-in, most sites will not want Static CDN Filters to be applied and static resource URLs to be rewritten." to "Most sites will NOT want Static CDN Filters applied when a user is logged-in. Content for logged-in users is often very dynamic and rewriting static resource URLs for logged-in users can cause problems. If you know that your logged-in user content is very static, you can enable Static CDN Filters for logged-in users to improve performance." (@jaswsinc can you confirm this wording makes sense?)
- [ ] In the dropdown options, I would change "(default option)" to "(recommended)"
- [ ] In the dropdown options, I would change "Yes, allow Static CDN Filters to be applied to logged-in users." to "Yes, enable Static CDN Filters for logged-in users."

jaswsinc:
> Change "When a user is logged-in, most sites will not want Static CDN Filters to be applied and static resource URLs to be rewritten." to "Most sites will NOT want Static CDN Filters applied when a user is logged-in. Content for logged-in users is often very dynamic and rewriting static resource URLs for logged-in users can cause problems. If you know that your logged-in user content is very static, you can enable Static CDN Filters for logged-in users to improve performance." (@jaswsinc can you confirm this wording makes sense?)

I can't imagine something being "very static" vs. "static". If it's a `.js`, `.css`, image, or some other static content, then there is no problem with it being served from the CDN at all times, even for logged-in users. 

So while the default setting should be to disable when logged-in, that's only because it is what we feel is easier to comprehend, given the context in which this is applied (or not applied) whenever you are logged in as an administrator; i.e., if logged-in user caching is off, one could argue that you might expect other parts of ZenCache (not just the caching engine) to be off also.

However, the truth is that this option should always be on as a good recommendation. Once you understand that it's going to be on for logged-in users, and that's perfectly OK, it is something that you will expect to see whenever you're logged in and testing your site.

Static is static. Static can be served via a CDN for improved performance, no matter.

kristineds:
@jaswsinc @raamdev Updated code. https://github.com/websharks/zencache-pro/commit/72958ec284fb4d263236266ed06c38c6206e0970 :+1: 

![Updated Text for UI Option Static CDN filter]
(https://cloud.githubusercontent.com/assets/7514953/11449522/f512e730-95b3-11e5-9676-79f978cab888.png)

raamdev:
@jaswsinc writes...
> I can't imagine something being "very static" vs. "static". If it's a .js, .css, image, or some other static content, then there is no problem with it being served from the CDN at all times, even for logged-in users.

Ah, right. When I wrote that I wasn't thinking only JavaScript/CSS--I was thinking PHP-generated content too (i.e., what is affected by Logged-In User Caching). I forgot for a moment that the context for that description was limited to Static CDN Filters, which only applies to static content.

> So while the default setting should be to disable when logged-in, that's only because it is what we feel is easier to comprehend, given the context in which this is applied (or not applied) whenever you are logged in as an administrator; i.e., if logged-in user caching is off, one could argue that you might expect other parts of ZenCache (not just the caching engine) to be off also.

Right. It's definitely easier to comprehend and leads to less confusion and fewer questions. If you specifically turn off Logged-In User Caching (or you leave it disabled as the default is), and you then discover that ZenCache is still caching _something_ (i.e., the Static CDN Filters are being applied) for Logged-In Users, your first reaction is to think that ZenCache is broken and not doing as you told it. 

Leaving Static CDN Filters disabled for Logged-In Users by default (just like caching for Logged-In Users) makes everything a lot more clear.

> However, the truth is that this option should always be on as a good recommendation. Once you understand that it's going to be on for logged-in users, and that's perfectly OK, it is something that you will expect to see whenever you're logged in and testing your site.

Right, I agree. That's really an education issue and not something easily fixed for first-time users or users who are not interested in learning, so I think keeping it disabled by default is still the best choice here.

**Maybe we could** leave things as they are now (Static CDN Filters for Logged-In Users disabled by default) but change our "(recommended)" string to the "Yes, enabled Static CDN Filters for Logged-In Users". That way if a site owner enables Logged-In User Caching and then looks at the related options for Static CDN Filters, they will see that we recommend enabling Static CDN Filters for Logged-In Users.

jaswsinc:
> Right, I agree. That's really an education issue and not something easily fixed for first-time users or users who are not interested in learning, so I think keeping it disabled by default is still the best choice here.

Yeah. You convinced me also. I agree.

> Maybe we could leave things as they are now (Static CDN Filters for Logged-In Users disabled by default) but change our "(recommended)" string to the "Yes, enabled Static CDN Filters for Logged-In Users". That way if a site owner enables Logged-In User Caching and then looks at the related options for Static CDN Filters, they will see that we recommend enabling Static CDN Filters for Logged-In Users.

Agree. The `(recommended)` label should be on the `Yes` option.

jaswsinc:
@kristineds If you read over the last two comments you will see there are some minor tweaks needed in this UI that you're working on here. If anything in the above is not clear enough just let us know. Thanks for your patience here. :-)

kristineds:
@jaswsinc @raamdev Like this? https://github.com/websharks/zencache-pro/commit/4b7c859dcf3e8df05961d1634b07e2a315ca6a3c

ZC Static CDN Filters - Yes recommended

jaswsinc:
@kristineds Yes, that's right. Now if you can alter the text above that option so that it reflects what we recommend as the best option that'd be great. Currently, there is conflicting information there. The description suggests that we recommend `No`, while the dropdown recommends `Yes`.

e.g., _This should almost always be set to `No`_ is incorrect now.

kristineds:
> This should almost always be set to `No` is incorrect now.

@jaswsinc @raamdev I removed that line and altered the description a bit. Let me know if you want to revise this further. ;)  https://github.com/websharks/zencache-pro/commit/a38aee5da6410f70694a91b28a6317e2135bf09a 

ZC UI Option enable static cdn filter for logged-in users

jaswsinc:
@kristineds Still conflicting information there in my view. See discussion here: https://github.com/websharks/zencache-pro/pull/189#issuecomment-160227925

kristineds:
@jaswsinc @raamdev How about this? https://github.com/websharks/zencache-pro/commit/cf05d9d03fe3695b6bc9f275f48729e6b1676803

screen shot 2015-12-02 at 10 35 07 pm

jaswsinc:
#### @kristineds @raamdev
Suggested description. Please review...

---

While this defaults to a value of `No`, it should almost always be set to `Yes`. Once you understand that Static CDN Filters can be applied globally—to speed things up for all visitors (logged-in or not logged-in), please choose `Yes` in the dropdown below.

**Note:** Static CDN Filters serve _static_ resources. Static resources, are, simply put, static. Thus, it is not a problem for any visitor (logged-in or not logged-in). To avoid confusion, this defaults to a value of `No`, and we ask that you set it to `Yes` on your own so that you'll know to expect this behavior; i.e., that static resources will always be served from the CDN, logged-in or not.

jaswsinc:
> How about this? cf05d9d

I still feel like that is confusing. I provided a recommendation above.

raamdev:
@jaswsinc Tweaked the messaging a bit for clarity; thoughts?

---

While this defaults to a value of `No`, it should almost always be set to `Yes`. This value defaults to `No` only because Logged-In User caching (see above) defaults to `No` and setting this value to `Yes` by default can cause confusion for some users. Once you understand that Static CDN Filters can be applied safely for all visitors (logged-in or not logged-in), please choose `Yes` in the dropdown below. If you are not using Static CDN Filters, the value below is ignored.

**Note:** Static CDN Filters serve _static_ resources. Static resources, are, simply put, static. Thus, it is not a problem to cache these resources for any visitor (logged-in or not logged-in). To avoid confusion, this defaults to a value of `No`, and we ask that you set it to `Yes` on your own so that you'll know to expect this behavior; i.e., that static resources will always be served from the CDN (logged-in or not logged-in) even though Logged-In User caching may be disabled above.

jaswsinc:
Looks great to me. Nice enhancements.

raamdev:
@kristineds Can you please update the description as per my last comment (https://github.com/websharks/zencache-pro/pull/189#issuecomment-161443599)? Please use a blue box for the "Note" paragraph and place that _after_ the dropdown box. Thank you. :smile:

kristineds:
@raamdev @jaswsinc Here you go. Let me know if I missed anything. https://github.com/websharks/zencache-pro/commit/86dad7c00bca555bb409892a48746c6f8ff7fa91

screen shot 2015-12-03 at 5 12 36 pm

raamdev:
@kristineds Sweeet! Thank you. I will merge this soon. :smile:

raamdev:
@kristineds Merged. Beautiful work! Thank you. :smile:

PR #191, opened 2015-12-02
Clear WooCommerce Product cache file when product stock changes.

raamdev:
See websharks/zencache#597


raamdev:
@jaswsinc I tested this and confirmed it's working as expected, but a quick code review before I merge this would be appreciated. :-)

jaswsinc:
@raamdev Awesome!!

Only thing I noticed was `empty($args['is_preview'])` seems out of place.
I think `$args` might be a mistake?

raamdev:
@jaswsinc writes...
> Only thing I noticed was empty($args['is_preview']) seems out of place.

Ah, good catch. Yep, that's a remanent of the code I build this from. Thanks! :-) Fixed in https://github.com/websharks/zencache-pro/commit/9c1de18cb62bcf63c8c87a06038b83a7e8622f42.

jaswsinc:
:+1:

PR #192, opened 2015-12-03
PR: feature/615

jaswsinc:
See: websharks/zencache#615
And discussion here: https://github.com/websharks/zencache/issues/618#issuecomment-159088528


jaswsinc:
### Summary of Changes

#### New Config. Option on Multisite Networks

This now defaults to being off; i.e., by default, child blogs are not checked any longer.



You can turn this back on by selecting the second option seen here.



---

#### Improved diagnostic reporting for sitemap-related errors.



---

#### Diagnostic reporting is now always disabled for child blogs.

i.e., ZenCache will only complain if the error is related to a primary sitemap location.

jaswsinc:
@raamdev Please review when you have some time. Thanks!

jaswsinc:
**Tip:** The easiest way to test this is to enter an invalid sitemap URL.

Then run the CRON manually: https://YOURSITE/?zencache_auto_cache_cron=1

raamdev:
@jaswsinc writes...
> Diagnostic reporting is now always disabled for child blogs.

If a Multisite Network admin _wants_ to see XML Sitemap errors for child blogs (when the Auto-Cache Engine is configured to check those), will they see them in `wp-content/cache/zencache/zencache-auto-cache.log` or is there no way to see those errors?

jaswsinc:
You won't see the diagnostic report in the log file. You will only see that specific URLs are not being hit if the sitemap for a specific child blog were being loaded. In other words, you would need to learn from what you don't find in that log file.

raamdev:
Copy that. Thank you.

PR #193, opened 2015-12-03
Improving XML Sitemap error checking.

raamdev:
See websharks/zencache#616

### DIFF
--- a/src/includes/classes/Actions.php
+++ b/src/includes/classes/Actions.php
@@ -576,6 +576,9 @@ class Actions extends AbsBase
             if (!($add_advanced_cache = $this->plugin->addAdvancedCache())) {
                 $query_args[GLOBAL_NS.'_advanced_cache_add_failure'] = $add_advanced_cache === null ? 'advanced-cache' : '1';
             }
+            if (!$this->plugin->options['auto_cache_enable'] || !$this->plugin->options['auto_cache_sitemap_url']) {
+                $this->plugin->autoCacheMaybeClearPrimaryXmlSitemapError(true);
+            }
             $this->plugin->updateBlogPaths(); // Multisite networks only.
         } else {
             if (!($remove_wp_cache_from_wp_config = $this->plugin->removeWpCacheFromWpConfig())) {
@@ -587,6 +590,7 @@ class Actions extends AbsBase
             if (!($remove_advanced_cache = $this->plugin->removeAdvancedCache())) {
                 $query_args[GLOBAL_NS.'_advanced_cache_remove_failure'] = '1';
             }
+            $this->plugin->autoCacheMaybeClearPrimaryXmlSitemapError(true);
         }
         $redirect_to = add_query_arg(urlencode_deep($query_args), $redirect_to);
 
--- a/src/includes/classes/AutoCache.php
+++ b/src/includes/classes/AutoCache.php
@@ -255,23 +255,10 @@ class AutoCache extends AbsBase
         if (!($sitemap = trim((string) $sitemap))) {
             goto finale; // Nothing we can do.
         }
-        if (is_wp_error($head = wp_remote_head($sitemap, array('redirection' => 5)))) {
-            $failure = 'WP_Http says: '.$head->get_error_message().'.';
-        } elseif (empty($head['response']['code']) || (int) $head['response']['code'] >= 400) {
-            $failure = sprintf(__('HEAD response code (<code>%1$s</code>) indicates an error.', SLUG_TD), esc_html((int) @$head['response']['code']));
-        } elseif (empty($head['headers']['content-type']) || stripos($head['headers']['content-type'], 'xml') === false) {
-            $failure = sprintf(__('Content-Type (<code>%1$s</code>) indicates an error.', SLUG_TD), esc_html((string) @$head['headers']['content-type']));
+        if (!$this->checkXmlSitemap($sitemap, $___recursive)) {
+            goto finale; // Nothing we can do.
         }
-        if ($failure) { // Failure encountered above? If so, skip to finale after possible notice to site owner.
-            if (!$this->is_child_blog && !$___recursive) { // If this is a primary sitemap location.
-                $this->plugin->enqueueMainNotice(
-                    sprintf(__('<strong>%1$s says...</strong> The Auto-Cache Engine is currently configured with an XML Sitemap location that could not be found. We suggest that you install the <a href="http://zencache.com/r/google-xml-sitemaps-plugin/" target="_blank">Google XML Sitemaps</a> plugin. Or, empty the XML Sitemap field and only use the list of URLs instead. See: <strong>Dashboard → %1$s → Auto-Cache Engine → XML Sitemap URL</strong>', SLUG_TD), esc_html(NAME)).'</p><hr />'.
-                    sprintf(__('<p><strong>Problematic Sitemap URL:</strong> <a href="%1$s" target="_blank">%1$s</a> / <strong>Diagnostic Report:</strong> %2$s', SLUG_TD), esc_html($sitemap), $failure),
-                    array('class' => 'error', 'persistent_key' => 'xml_sitemap_missing')
-                );
-            }
-            goto finale; // Nothing more we can do in this case.
-        }
+
         if ($xml_reader->open($sitemap)) {
             while ($xml_reader->read()) {
                 if ($xml_reader->nodeType === $xml_reader::ELEMENT) {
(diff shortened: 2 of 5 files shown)


raamdev:
@jaswsinc Submitting this for review and feedback. I was just working on `maybeClearPrimaryXmlSitemapError()` and was thinking that we could hook that onto `admin_init`, however then I realized that the `__construct()` for that class calls `run()` to run the Auto-Cache Engine, so that `maybeClearPrimaryXmlSitemapError()` would either need to be moved to a separate file, or we need to modify the `__construct()` and anything in the codebase that calls the Auto-Cache Engine.

jaswsinc:
- I like the separate `checkXmlSitemap()` function. That's nice.
- I think you might not need `maybeClearPrimaryXmlSitemapError()`.

  Instead, you already have [these lines](https://github.com/websharks/zencache-pro/blob/c7881f4eaf6ac8b9279e8b09dd8bc43174c57a90/src/includes/classes/AutoCache.php#L397-L399) in place which seem logical to me. That's going to be capable of picking up valid sitemaps that did not result in an error. It will pick those on each and every run of the CRON job, so that seems like a good place to detect whether it is now valid or not.

  What's missing (at the moment), is a way to correlate a given notice with a specific sitemap URL; i.e., which sitemap is now valid, and which ones are not? Right now, we have `'persistent_key' => 'xml_sitemap_missing'`. You might give some thought to making that unique in some way, so that you can identify it by URL.

  For instance, maybe the persistent notice key could be a hash of the URL.

  ```
  'persistent_key' => 'xml_sitemap_missing'.sha1($sitemap)
  ```

  So whenever you dismiss, you can dismiss it by URL? Not sure.

  ```
  $this->plugin->dismissMainNotice('xml_sitemap_missing'.sha1($sitemap));
  ```

raamdev:
> you already have these lines in place which seem logical to me.

That only runs when the Auto-Cache Engine runs. The problem I'm trying to solve is getting rid of the notice as soon as the issue with the XML Sitemap has been resolved, i.e., not after 15 minutes but right away (e.g., after the site owner installs an XML Sitemap plugin and the sitemap becomes accessible).

raamdev:
> What's missing (at the moment), is a way to correlate a given notice with a specific sitemap URL; i.e., which sitemap is now valid, and which ones are not?

We don't report errors for Child Blogs, so why would there be more than one Primary Sitemap URL?

jaswsinc:
> after the site owner installs an XML Sitemap plugin and the sitemap becomes accessible).

I see. Then I'd agree with what you alluded to before. Either move this out of the ACE class and instead create a new class member as a `Plugin/Closure/*` somewhere that is attached to `admin_init`. You can run a check and then call the dismiss handler to remove the notice.

```php
$this->plugin->dismissMainNotice('xml_sitemap_missing');
```

If you can find a hook that would detect any new plugin being installed, or that a sitemap plugin was installed, that might be enough. I'd try to refrain from pinging the invalid sitemap URL over and over again if you can avoid it, but you might not have a choice given the mission that you're on here.

If you'd rather keep everything in the ACE class file, keep in mind that a class can have both OOP and also static methods. So if you want to dodge the `__construct()` method create some of the class members in that file as `public static function`s. That way you can then call them stand-alone.

```php
AutoCache:: checkXmlSitemap()
```

http://php.net/manual/en/language.oop5.static.php

jaswsinc:
> so why would there be more than one Primary Sitemap URL?

It's not likely, given what you just mentioned. However, if a site owner has changed the `/sitemap.xml` path to something different more than once or twice, there could be errors triggered for each of the paths they gave to ZenCache. That's an edge case though, you're right.

raamdev:
> If you can find a hook that would detect any new plugin being installed, or that a sitemap plugin was installed, that might be enough. I'd try to refrain from pinging the invalid sitemap URL over and over again if you can avoid it, but you might not have a choice given the mission that you're on here.

I thought about that, but a missing XML Sitemap plugin is just one of several possible reasons that site owner might see the XML Sitemap error, and all the _other_ reasons are exactly when an intelligent notice would be particularly useful (e.g., if it's a 404 error and there _is_ an XML Sitemap plugin installed and active, the troubleshooting that goes into figuring out what's causing the problem would be greatly benefited by ZenCache getting rid of that error messages immediately). 

My feeling is that running the check on every `admin_init` and pinging the invalid sitemap URL will be fine because that error message will be persistent. If a site owner wants to get rid of the error message (and stop ZenCache from pinging that invalid URL) they either need to fix the problem, clear the XML Sitemap field, or disable the Auto-Cache Engine altogether. 

> if a site owner has changed the /sitemap.xml path to something different more than once or twice, there could be errors triggered for each of the paths they gave to ZenCache.

Oh, right. Great catch! I keep forgetting that the SHA1 of actual notice itself is stored to prevent dupes.

Yeah, I'll have to figure out how to avoid that issue.

raamdev:
@jaswsinc writes...
> if a site owner has changed the /sitemap.xml path to something different more than once or twice, there could be errors triggered for each of the paths they gave to ZenCache.

Actually, after reviewing `NoticeUtils.php`, this doesn't seem like something that I need to worry about. 

If we can assume that we're only concerned with a single Primary Site XML Sitemap URL at a time, i.e., a site is not going to have more than one "Primary Site XML Sitemap URL" at a given time, then I could just run `$this->plugin->dismissMainNotice('xml_sitemap_missing');` to clear any previous notice that may be referencing a different Sitemap URL _before_ enqueuing the current error notice.

Here's why that would work: When we're dismissing a persistent notice with `$this->plugin->dismissMainNotice()` [any notices matching that `persistent_key`, _regardless_ of the unique SHA1 of the notice](https://github.com/websharks/zencache-pro/blob/151114/src/includes/closures/Plugin/NoticeUtils.php#L65-L66), are dismissed.

Here's what would happen: 

1. Persistent Error Notice with `example.com/sitemap.xml` enqueued and shown to site owner.
1. Site owner changes Sitemap URL to `example.com/my-sitemap.xml`
1. ZenCache finds that `example.com/my-sitemap.xml` is also invalid
1. ZenCache runs `$this->plugin->dismissMainNotice('xml_sitemap_missing');`, [which deletes _any and all_ notices](https://github.com/websharks/zencache-pro/blob/151114/src/includes/closures/Plugin/NoticeUtils.php#L65-L66) associated with the persistent key `xml_sitemap_missing`, including the one mentioning `example.com/sitemap.xml`.
1. ZenCache enqueues a new notice indicating that `example.com/my-sitemap.xml` is invalid.

---

Now to test this theory. :smile:

raamdev:
@jaswsinc Please see https://github.com/websharks/zencache-pro/commit/95eeabe583dce0f0e7b01e7e00fea8db73382817. I tested this and it's working as expected. It took some trial-and-error to address all the various scenarios, but I think I got them all. 

I was originally going to create a new `closures/Plugin/AutoCacheNoticeUtils.php` file, but I realized that the stuff I was putting there could just as easily go in the existing `closures/Plugin/AutoCacheUtils.php` file, so that's what I did.

Please see the DocBlocks (and the `@note`s) for further details.

In my tests, the Dashboard notice is now fully responsive. 

- If you enable an XML Sitemap plugin to fix the problem, the error notice goes away as soon as the plugin is active.
- If you disable ZenCache, the Sitemap error notice goes away.
- If you disable the Auto-Cache Engine, the Sitemap error notice goes away.
- If you clear the XML Sitemap field inside the Auto-Cache Engine options and then save the settings, the Sitemap error notice goes away.

---

The only issue that remains is that the Error Notice that appears shows a "dismiss" link. Clicking the link doesn't get rid of the notice (because it's intentionally persistent, as this check occurs on every `admin_init`). I believe fixing this issue will require updating `NoticeUtils.php`. E.g., maybe we can create a new `persistent_not_dismissable_key` that can be used in place of `persistent_key` for situations like these; see [these lines](https://github.com/websharks/zencache-pro/blob/000000-dev/src/includes/closures/Plugin/NoticeUtils.php#L149-L152).

jaswsinc:
Beautiful work. Fantastic!

> maybe we can create a new persistent_not_dismissable_key that can be used in place of persistent_key for situations like these; see these lines.

That's an awesome idea. Or maybe an attribute like `dismissable` [here](https://github.com/websharks/zencache-pro/blob/000000-dev/src/includes/closures/Plugin/NoticeUtils.php#L260), and then change [the lines](https://github.com/websharks/zencache-pro/blob/000000-dev/src/includes/closures/Plugin/NoticeUtils.php#L149-L152) you referenced so that they will recognize that new behavioral attribute; i.e., a notice can be persistent with a key, but also dismissable or not dismissable.

raamdev:
> Or maybe an attribute like dismissable here, and then change the lines you referenced so that they will recognize that new behavioral attribute; i.e., a notice can be persistent with a key, but also dismissable or not dismissable.

That's even better! Awesome. Wow, I'm beginning to appreciate the flexibility of those notice utils. :smile:

raamdev:
@jaswsinc ↑ Committed changes you suggested and updated (and tested) the improvement to Notice Utils.

Also, note my response regarding `$__recursive` here: https://github.com/websharks/zencache-pro/pull/193#commitcomment-14776755

jaswsinc:
> and updated (and tested) the improvement to Notice Utils.

Cool! That looks great to me.

I left you note here: https://github.com/websharks/zencache-pro/commit/95eeabe583dce0f0e7b01e7e00fea8db73382817#commitcomment-14777815

PR #194, opened 2015-12-04
Disable CDN Filters when ZENCACHE_ALLOWED is false or DONOTCACHEPAGE set

raamdev:
See websharks/zencache#628

### DIFF
--- a/src/includes/classes/CdnFilters.php
+++ b/src/includes/classes/CdnFilters.php
@@ -216,6 +216,18 @@ class CdnFilters extends AbsBase
         if (is_admin()) {
             return; // Not applicable.
         }
+        if (defined('ZENCACHE_ALLOWED') && !ZENCACHE_ALLOWED) {
+            return; // Disable in this case.
+        }
+        if (isset($_SERVER['ZENCACHE_ALLOWED']) && !$_SERVER['ZENCACHE_ALLOWED']) {
+            return; // Disable in this case.
+        }
+        if (defined('DONOTCACHEPAGE')) {
+            return; // Disable in this case.
+        }
+        if (isset($_SERVER['DONOTCACHEPAGE'])) {
+            return; // Disable in this case.
+        }
         if (!$this->cdn_enable) {
             return; // Disabled currently.
         }


raamdev:
@jaswsinc A quick review here, please. 

https://github.com/websharks/zencache/issues/628 mentions adding a new `ZENCACHE_CDN_ALLOWED` and `ZENCACHE_HTMLC_ALLOWED` (which I suggested to Renz would be a good idea), but reviewing the code and thinking about it some more, I'm thinking what I've done in this PR is probably good enough. Those two additional constants likely won't be used very much and it would be hard to clarify how exactly they work (e.g., in what sorts of scenarios will setting those constants overwrite the current behavior, or in what contexts do they work?).

Any thoughts on this?

jaswsinc:
A few things come to mind.

- Static CDN Filters are applied at runtime in a somewhat random fashion; i.e., whenever URLs are being built by a template or by WordPress itself, a plugin, etc. So this doesn't happen in the shutdown phase like our caching engine does, it happens during the regular flow of events.

  Therefore, trying to read `ZENCACHE_ALLOWED` or `DONOTCACHEPAGE` in the Static CDN class may or may not work, depending on when these constants were defined exactly.

- We have [an open issue](https://github.com/websharks/zencache/issues/535) that is a request for Static CDN Filters to be applied during the shutdown phase instead, which would change this and make it easier for `ZENCACHE_ALLOWED` or `DONOTCACHEPAGE` to work as you expect. So it might be worth taking a closer look at that now.

  However, I remember looking over that issue and it gave me pause, because applying Static CDN Filters during the shutdown phase would add some additional parsing complexity and have a slightly negative impact on performance also. The way they are applied now (i.e., as-is), is very simple and easy for us to maintain. No chance of odd conflicts and/or any parsing peculiarities.

- Given both of those challenges, I'm leaning in the direction of there being a `ZENCACHE_STATIC_CDN_FILTERS_DISABLE` constant (i.e., removing the `_ALLOWED` keyword to distinguish this from others). The only way this new constant would work though, is if you set it in your `/wp-config.php` file, or if you define it before the filters are setup with a hook like:

  ```php
  <?php
  add_action('before_zencache_setup', function() {
      define('ZENCACHE_STATIC_CDN_FILTERS_DISABLE', true);
  }, -PHP_INT_MAX);
  ```

  So _that's_ what we could document as a recommendation. However, I'd still leave what you have also. No harm in trying to read `ZENCACHE_ALLOWED` and/or `DONOTCACHEPAGE` if those are already set and we can detect them early enough for them to make a difference.

raamdev:
@jaswsinc writes...
> Given both of those challenges, I'm leaning in the direction of there being a ZENCACHE_STATIC_CDN_FILTERS_DISABLE constant

So are you suggesting that we implement that now (without making Static CDN Filters apply during the shutdown phase), the same way `ZENCACHE_ALLOWED` is implemented, except make it specific to CDN Filters and then write a specific recommendation (KB Article) that explains if you want to disable CDN Filters, you must do it in `wp-config.php` or `before_zencache_setup` (as you gave the example above)?

jaswsinc:
That's what I was suggesting in that last point, yes. However, I'm not sure if that's the right way to go about this or not. I was just reading back over the post from Renz, where he writes...

> Setting define('ZENCACHE_ALLOWED', FALSE); or define('DONOTCACHEPAGE', TRUE); should disable the ZenCache Static CDN Filters as well as the caching itself. There should probably be a separate constant to allow for specifically disabling Static CDN Filters and the HTML Compressor (e.g., ZENCACHE_CDN_ALLOWED and ZENCACHE_HTMLC_ALLOWED).

So it seems the only reason we are working on this in the first place, is to avoid confusion; i.e., to make it so that if `ZENCACHE_ALLOWED` or `DONOTCACHEPAGE` are set, then Static CDN Filters are disabled too. Do I have that right? Adding a new constant won't accomplish this, it may only confuse things further. I think that's what you were trying to illustrate in your notes above also.

The only way that I see for us to accomplish this perfectly, would be for Static CDN Filters to become a part of the caching routine; i.e., be performed in the shutdown phase, so that everything could be controlled by `ZENCACHE_ALLOWED` or `DONOTCACHEPAGE`.

On the other hand, I'm not convinced that mixing Static CDN Filters with the cache engine itself is a good idea. In most use cases, these really are two separate functionalities. They are both offered by the same plugin—ZenCache. But strictly speaking, these are two different things.

1. Cache page content output by WordPress (primary functionality).
2. Filter `home_url()` (and the like), so the serves static content from a CDN.

raamdev:
> So it seems the only reason we are working on this in the first place, is to avoid confusion; i.e., to make it so that if ZENCACHE_ALLOWED or DONOTCACHEPAGE are set, then Static CDN Filters are disabled too. Do I have that right?

Well, not entirely. See https://websharks.zendesk.com/agent/tickets/9784 for some background.

raamdev:
> So it seems the only reason we are working on this in the first place, is to avoid confusion;

If someone sets `ZENCACHE_ALLOWED` to false, i.e., ZenCache is not allowed on a particular page, then Static CDN Filters (i.e., part of ZenCache) should not be allowed either. That logic would be the assumed behavior, especially given the name of that constant: `ZENCACHE_ALLOWED`. So yes, part the problem is confusion around why Static CDN Filters are being applied when you've told ZenCache that it is now allowed.

However, barring an easy way to make Static CDN Filters obey `ZENCACHE_ALLOWED`, adding another constant, e.g., `ZENCACHE_STATIC_CDN_FILTERS_DISABLE`, that can be used (perhaps in conjunction with `ZENCACHE_ALLOWED`) to specifically disable the Static CDN Filters on certain pages would be sufficient. A KB Article that explains why `ZENCACHE_ALLOWED` cannot be used to reliably disable Static CDN Filters would, in my mind, be enough to clarify things for any site owners who may run into this scenario.

jaswsinc:
Thank you..

A user writes...

> I need to disable CDN support (and probably Zencache itself) for some specific pages - I know, I can use define('ZENCACHE_ALLOWED', FALSE); to disable Zencache, but this still uses static CDN if globally enabled.

Here's what is possible in the current version of ZC Pro.

```php
<?php
add_filter('zencache_options', function(options) {
    if(is_page('my-page')) {
        $options['cdn_enable'] = '0';
    }
    return $options;
});
```

jaswsinc:
> So yes, part the problem is confusion around why Static CDN Filters are being applied when you've told ZenCache that it is now allowed.

Copy. Yeah. I think the changes you've already made, plus a new KBA, and maybe the code sample I just gave would be enough. In the meantime, we can add another check in the plus column for [this open issue](https://github.com/websharks/zencache/issues/535) and perhaps implement that after all.

raamdev:
> I think the changes you've already made, plus a new KBA, and maybe the code sample I just gave would be enough.

Roger that. Thank you. :-) 

> In the meantime, we can add another check in the plus column for this open issue and perhaps implement that after all.

Yep, that work would definitely make the Static CDN Filters have a much greater "reach" and be able to catch more static URLs (which is a primary reason many ZenCache Pro site owners use CDN Linker in favor of the Static CDN Filters).

PR #196, opened 2015-12-06
Set a $network_wide default in case parameter is not passed to hook

raamdev:
See websharks/zencache#631

### DIFF
--- a/src/includes/classes/Plugin.php
+++ b/src/includes/classes/Plugin.php
@@ -469,8 +469,8 @@ class Plugin extends AbsBaseAp
 
         add_filter('enable_live_network_counts', array($this, 'updateBlogPaths'));
 
-        add_action('activated_plugin', array($this, 'autoClearOnPluginActivationDeactivation'));
-        add_action('deactivated_plugin', array($this, 'autoClearOnPluginActivationDeactivation'));
+        add_action('activated_plugin', array($this, 'autoClearOnPluginActivationDeactivation'), 10, 2);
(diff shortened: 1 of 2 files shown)


raamdev:
@jaswsinc It looks like the issue was that the `$network_wide` parameter was not always being passed by the `plugin_activated` and `plugin_deactivated` hooks. I was able to reproduce the issue in my test environment consistently (occurred on every activate/deactivate of ZenCache Pro), and the change in this PR fixed the problem. Any objections here?

raamdev:
@jaswsinc Also note that `$network_wide` defaults to false in WP Core: https://github.com/WordPress/WordPress/blob/60c8e272f50e5c34d9cf592d3440434aefd3fad5/wp-admin/includes/plugin.php#L518

jaswsinc:
> I was able to reproduce the issue in my test environment consistently (occurred on every activate/deactivate of ZenCache Pro), and the change in this PR fixed the problem. Any objections here?

Nope, that looks good to me. Probably a good idea to make that an optional parameter anyway. However, I just noticed why we aren't getting the second argument there. We didn't ask for it. That's probably my mistake. See: https://github.com/websharks/zencache-pro/blob/6e5a0e642ea5e264ac2599807ba9558835cf1a9e/src/includes/classes/Plugin.php#L472-L473

raamdev:
@jaswsinc Great catch! That would explain it! I was a little puzzled why that second parameter would not _always_ be passed after reviewing WP Core code. I've fixed this in https://github.com/websharks/zencache-pro/commit/1fdf6c8869cd456914dd4eb20dcc7d7f4431b515

PR #198, opened 2015-12-15
PR: feature/624

jaswsinc:
See: websharks/zencache#624


jaswsinc:
## Summary of Changes

- The new default value for `cache_clear_opcache_enable` is now `0` (off), and the wording and suggestions offered in the Dashboard for this have been updated to mention that clearing the Opcache is *NOT* necessary.

- Why? It is no longer necessary, because now we automatically invalidate the `advanced-cache.php` (regardless of configuration), which allows ZenCache to function as expected whenever you update your configuration options. That was the most important reason for needing to clear the opcode cache in previous versions.

  The reason it must be invalidated, is because the `wp-content/advanced-cache.php` dropin is rebuilt each time you update your ZC config options, and so that file needs to be invalidated anytime it changes. Otherwise, you could update your ZC options and nothing would actually change—at least not until your opcache cache expires naturally, which can vary wildly depending on the server config.

- Clearing the entire Opcache is now 100% opt-in. You can enable it from the Manual Cache Clearing section, or choose to clear the entire opcode cache from options in the admin bar. All this means is that `cache_clear_opcache_enable` now defaults to a value of `0` as described above.

jaswsinc:
@raamdev Please review when you get some time. Thanks! :-)

raamdev:
@jaswsinc Beautiful! I love that we're not automatically clearing the entire Opcache every time now; that seems much safer and less wasteful overall. :-)

raamdev:
Reviewed and merged. Thank you!

PR #199, opened 2015-12-18
Run htaccess file through wp_check_invalid_utf8() before touching it.

raamdev:
See websharks/zencache#633


raamdev:
@jaswsinc Quick review here when you get a chance. :-) I ran this through a few tests and it seems to work fine.

Note the reason I added a `strlen()` check to the conditional: The `wp_check_invalid_utf8()` function runs the same check at the beginning and returns an empty string, which is the same thing it returns if there is invalid UTF-8.

PR #200, opened 2015-12-18
Acquire exclusive lock when reading and writing htaccess

raamdev:
See websharks/zencache#633

### DIFF
--- a/src/includes/classes/VsUpgrades.php
+++ b/src/includes/classes/VsUpgrades.php
@@ -330,7 +330,7 @@ class VsUpgrades extends AbsBase
             if (!($htaccess_file = $this->plugin->findHtaccessFile())) {
                 return; // File does not exist.
             }
-            if ($htaccess = $this->plugin>readHtaccessFile($htaccess_file, 'ZenCache')) {
+            if ($htaccess = $this->plugin->readHtaccessFile($htaccess_file, 'ZenCache')) {
 
                 if ($htaccess['marker_exists'] === false) {
                     flock($htaccess['fp'], LOCK_UN);
--- a/src/includes/classes/VsUpgrades.php
+++ b/src/includes/classes/VsUpgrades.php
@@ -333,15 +333,15 @@ class VsUpgrades extends AbsBase
             if ($htaccess = $this->plugin->readHtaccessFile($htaccess_file, 'ZenCache')) {
 
                 if ($htaccess['marker_exists'] === false) {
-                    flock($htaccess['fp'], LOCK_UN);
-                    fclose($htaccess['fp']);
+                    $this->plugin->closeHtaccessFile($htaccess); // No need to write to htaccess file in this case.
                     return; // Template blocks are already gone.
                 }
                 if (is_dir($templates_dir = dirname(dirname(__FILE__)).'/templates/htaccess/back-compat')) {
-                    $htaccess_file_contents = str_replace(file_get_contents($templates_dir.'/v151114.txt'), '', $htaccess['file_contents']);
-                    $htaccess_file_contents = str_replace(file_get_contents($templates_dir.'/v151114-2.txt'), '', $htaccess['file_contents']);
+                    $htaccess['file_contents'] = str_replace(file_get_contents($templates_dir.'/v151114.txt'), '', $htaccess['file_contents']);
+                    $htaccess['file_contents'] = str_replace(file_get_contents($templates_dir.'/v151114-2.txt'), '', $htaccess['file_contents']);
+                    $htaccess['file_contents'] = trim($htaccess['file_contents']);
 
-                    if (!$this->plugin->writeHtaccessFile($htaccess['fp'], $htaccess_file_contents, 'ZenCache', false)) {
+                    if (!$this->plugin->writeHtaccessFile($htaccess, false, 'ZenCache')) {
                         return; // Failure; could not write changes.
                     }
                 }
--- a/src/includes/classes/VsUpgrades.php
+++ b/src/includes/classes/VsUpgrades.php
@@ -330,12 +330,10 @@ class VsUpgrades extends AbsBase
             if (!($htaccess_file = $this->plugin->findHtaccessFile())) {
                 return; // File does not exist.
             }
-            if ($htaccess = $this->plugin->readHtaccessFile($htaccess_file, 'ZenCache')) {
-
-                if ($htaccess['marker_exists'] === false) {
-                    $this->plugin->closeHtaccessFile($htaccess); // No need to write to htaccess file in this case.
-                    return; // Template blocks are already gone.
-                }
+            if (!$this->plugin->findHtaccessMarker('ZenCache')) {
+                return; // Template blocks are already gone.
+            }
+            if ($htaccess = $this->plugin->readHtaccessFile($htaccess_file)) {
                 if (is_dir($templates_dir = dirname(dirname(__FILE__)).'/templates/htaccess/back-compat')) {
                     $htaccess['file_contents'] = str_replace(file_get_contents($templates_dir.'/v151114.txt'), '', $htaccess['file_contents']);
                     $htaccess['file_contents'] = str_replace(file_get_contents($templates_dir.'/v151114-2.txt'), '', $htaccess['file_contents']);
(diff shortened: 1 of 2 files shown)


raamdev:
@jaswsinc I tested this, but I'd love a review here.

raamdev:
@jaswsinc I've updated this to use a single exclusive lock when both reading _and_ writing the htaccess file. I tested this, but I'd appreciate a review and further testing before merging. Thanks!

raamdev:
@jaswsinc I think this is ready for a final review. If you're satisfied, I'll merge this into the RC branch. :-)

jaswsinc:
If I enable Static CDN Filters and save options.
Then disable Static CDN Filters and save options.

I end up with this in my `.htaccess` file.

```
# BEGIN ZenCache WmVuQ2FjaGU (the WmVuQ2FjaGU marker is required for ZenCache; do not remove)
# END ZenCache WmVuQ2FjaGU
```

Nothing wrong with that, but I would expect there to be nothing, since the comments are not doing anything. I think this occurs because the `addWpHtaccess()` runs through all the motions, but since Static CDN Filters are off (and that's the only `.htaccess` tweak we make at the moment), I end up with just the comments.

raamdev:
@jaswsinc: Hmm, it seems there's still something wonky: 

- Install v151114 on a clean WP install (no previous ZenCache options), activate, enable with Static CDN Filters. Old `.htaccess` rules are added (without the unique marker). 
- Now copy/paste the `zencache-pro` folder from a zip built using the `feature/633b` branch. When you reload the Dashboard it says " ZenCache detected a new version of itself. Recompiling...", however when you inspect your `.htaccess` you find this:

```
# BEGIN ZenCache WmVuQ2FjaGU (the WmVuQ2FjaGU marker is required for ZenCache; do not remove)
<IfModule headers_module>
  </FilesMatch>
</IfModule>
# END ZenCache WmVuQ2FjaGU

# BEGIN ZenCache
<IfModule headers_module>
  </FilesMatch>
</IfModule>
# END ZenCache
```

I.e., it's not properly removing the previous version of the template... Hmmm. 

I'm suspecting this might have something to do with my recent changes to the `VsUpgrades.php` file (where I updated the `fromLte151114()` routine to use the new htaccess utils).

I've gotta run but I wanted to post this here before I left so I know where I left off, or in case you happen to catch what I missed. :-)

jaswsinc:
Suggestions...

- I think the bug you reported above is caused by [these lines of code](https://github.com/websharks/zencache-pro/blob/feature/633b/src/includes/classes/VsUpgrades.php#L341-L342). Notice that `$htaccess_file_contents` receives two writes, but no reads. In both replacements you are operating on `$htaccess['file_contents']`.

- At the top of the `HtaccessUtils.php` file, you could add a class property so that every method can access it from one place.

  ```php
  /*
   * Unique comment marker.
   *
   * @since 15xxxx Enhancing `.htaccess` tweaks.
   *
   * @return string Used in `.htaccess` parsing.
   */
  $self->htaccess_marker = 'WmVuQ2FjaGU';
  ```

- At some point you might consider a little refactoring in `readHtaccessFile()`. Does that method need either of its existing two arguments, or can it work that out all by itself? If so, I think your method would be even easier to work with that way. I can just call `$self->readHtaccessFile()` without giving it any args. That would also make that method slightly more official; i.e., it will know everything about where to find and how to read the file all by itself.

  If you want to keep the arguments for added flexibility I think that's fine, but they could be optional; i.e., if they are not passed in, they could default to the most common use case for that method call.

raamdev:
@jaswsinc writes...
> I think the bug you reported above is caused by these lines of code.

Yep, you were right. Thanks for spotting that. :smile: 

> you could add a class property so that every method can access it from one place.

Done! Great idea. Thank you.

> At some point you might consider a little refactoring

Done. I refactored things in https://github.com/websharks/zencache-pro/commit/b5a4cb7883116ed4ca21d06d4a10f8bff1089db8 to make those methods more "official". 

---

I also updated things (see https://github.com/websharks/zencache-pro/commit/b5a4cb7883116ed4ca21d06d4a10f8bff1089db8#diff-9c436c451e7c22ab8953f50e98f9b793R52) so that nothing is added to the `.htaccess` file when we don't have a reason to, i.e., we no longer add an empty comment block.

---

I also got rid of an annoying extra newline that the `fromLte151114()` upgrade routine was leaving behind (see https://github.com/websharks/zencache-pro/commit/b5a4cb7883116ed4ca21d06d4a10f8bff1089db8#diff-31c11a6b07c64eae30145f4aaa131a47R342). 

---

I tested everything again, including the upgrade routine, and it's working great now. I'd love one more test from you before I publish a GA release with these changes.

jaswsinc:
Looks great! I was just poking around through the code and I have a few more comments.

- In the `readHtaccessFile()` method, I think the `DISALLOW_FILE_MODS` check should come before the automatic creation of the `.htaccess` file. If `DISALLOW_FILE_MODS` is true, we shouldn't create the file either, even if it's just a blank file.

- In `readHtaccessFile()`, would it be a good idea to call `rewind()` on the resource handle before we return it? i.e., just as a precaution, that method could rewind after it reads to make sure the pointer is always at the beginning.

- The `closeHtaccessFile()` method could include an `array` type hint in its signature. That way if an array is not passed it will trigger a fatal error; i.e., inform us that we messed up somewhere.

  ```php
  $self->closeHtaccessFile = function (array $htaccess) use ($self) {
  ```

  - Same thing in `writeHtaccessFile()`. Add the `array` type hint.

-  Nitpicking... you could shorten this a bit:

  ```php
  if (empty($htaccess_marker)) {
        $htaccess_marker = $self->htaccess_marker;
    }
  ```

  In honor of Elvis...

  ```php
  $htaccess_marker = $htaccess_marker ?: $self->htaccess_marker;
  ```

- For good measure, there should probably be a `DISALLOW_FILE_MODS` check in the `writeHtaccessFile()` method. That way if we attempt to write it will fail if we are not allowed to do so.

jaswsinc:
> I also updated things (see b5a4cb7#diff-9c436c451e7c22ab8953f50e98f9b793R52) so that nothing is added to the .htaccess file when we don't have a reason to, i.e., we no longer add an empty comment block.

Wonderful! Love that.

raamdev:
> I was just poking around through the code and I have a few more comments.

Awesome! Great suggestions. :smile: Updated in https://github.com/websharks/zencache-pro/commit/c2111b2e237234bcc438e26799f3292b7f99c763.

jaswsinc:
> I tested everything again, including the upgrade routine, and it's working great now. I'd love one more test from you before I publish a GA release with these changes.

> Awesome! Great suggestions.  Updated in c2111b2.

Fantastic. I'm running some tests now.

jaswsinc:
:100: :+1: Working great on my side.

raamdev:
@jaswsinc Woohoo! Thanks so much for all the help. :grin:

jaswsinc:
One more thing. I think `readHtaccessfile()` could do a better job of separating the difference between "reading" and "writing". We should make every effort to return true if the file is readable. Even if it's not writable or if `DISALLOW_FILE_MODS` is true, we can still "read" it.

Looking over that method again, it seems the intended behavior is to "open" (e.g., `openHtaccessFile()`) for reads and writes. That might be a simpler solution; i.e., just rename the method?

On the other hand, there are times when we _only_ need to read the file. For instance, in `removeHtaccessFile()` we can return true if the marker is not present, even if we are unable or not allowed to write to the file.

jaswsinc:
> Thanks so much for all the help.

Absolutely my pleasure :-)

raamdev:
@jaswsinc Yeah, but that would require refactoring `readHtaccessFile()` to include the `fopen()` mode as well...

i.e., is it `fopen($htaccess_file, 'rb+')` (read+write) or `fopen($htaccess_file, 'rb')` (read)?

When I was writing that method I realized there were _some_ limitations, but at the same time I was reminded that I'm working within the context of the `htaccess` utilities here... not the entire ZenCache plugin (i.e., these are `htaccess` utilities specifically). It seems there are currently so few scenarios where I need to _only_ read that creating more utilities methods is overkill.

That said, I totally agree with you that `removeWpHtaccess()` can (and should) return true if the marker is not present, even if we are unable or not allowed to write to the file.

raamdev:
@jaswsinc I separated the search for the marker into a separate `findHtaccessMarker()` that uses `file_get_contents()` and doesn't require an exclusive lock or write access. See https://github.com/websharks/zencache-pro/commit/2f3f7c398e85277007e39c9c5bfbb62ded3f1f49

jaswsinc:
Looks great to me. Nice!

raamdev:
@jaswsinc Caught a nasty little bug while testing. https://github.com/websharks/zencache-pro/commit/ceb325ecf8143d212251b4dbfcc17a262c2181dd

raamdev:
I ran this branch through several tests, including an upgrade from v151114 via the [overwrite folder method](https://github.com/websharks/zencache-pro/pull/200#issuecomment-166013126). Everything checks out good now.

PR #202, opened 2015-12-27
Fix bug that results in ZenCache not disabling Akismet comment nonce

raamdev:
See websharks/zencache#642

### DIFF
--- a/src/includes/classes/Plugin.php
+++ b/src/includes/classes/Plugin.php
@@ -519,7 +519,9 @@ class Plugin extends AbsBaseAp
 
         /*[pro strip-from="lite"]*/
         if ($this->options['enable'] && $this->applyWpFilters(GLOBAL_NS.'_disable_akismet_comment_nonce', true)) {
-            add_filter('akismet_comment_nonce', '__return_false'); // See: <http://jas.xyz/1R23f5c>
+            add_filter('akismet_comment_nonce', function() {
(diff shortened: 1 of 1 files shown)


raamdev:
@jaswsinc Quick review here when you get a chance. Also, is there a reason why this filter is being stripped from the Lite version of ZenCache (via ` /*[pro strip-from="lite"]*/`)? It seems that disabling the Akismet comment nonce would be equally applicable there...

jaswsinc:
@raamdev This looks great to me.

I read over the investigation that you did. Great catch! :smile: 

> Also, is there a reason why this filter is being stripped from the Lite version of ZenCache (via /*[pro strip-from="lite"]*/)? It seems that disabling the Akismet comment nonce would be equally applicable there...

I agree, it should be included in the lite version too; i.e., we should remove the markers.

I think the reason I was stripping it was because I had thought previously that this would impact logged-in user caching only, which is only available in the pro version. However, now that I'm looking at this again, that particular aspect will impact _any_ user, not just those that are logged in.

In other areas, the conditionals that we use already consider both lite and pro, so that marker removal should do the same. Referencing: https://github.com/websharks/zencache-pro/blob/802c8867d3f414307dce96a7f3757d3b737c481e/src/includes/closures/Ac/ObUtils.php#L274-L280

In summary, we just need to remove the `/*[pro strip-from="lite"]*/` marker so that ZC will automatically disable the Akismet nonce whenever ZC is enabled for a site.

PR #203, opened 2015-12-27
Add option to cache nonce values

raamdev:
See websharks/zencache#637

### DIFF
--- a/src/includes/classes/Plugin.php
+++ b/src/includes/classes/Plugin.php
@@ -316,6 +316,7 @@ class Plugin extends AbsBaseAp
             'get_requests'        => '0', // `0|1`.
             'feeds_enable'        => '0', // `0|1`.
             'cache_404_requests'  => '0', // `0|1`.
+            'cache_nonce_values'  => '0', // `0|1`.
 
             /* Related to exclusions. */
 
--- a/src/includes/classes/Plugin.php
+++ b/src/includes/classes/Plugin.php
@@ -311,12 +311,13 @@ class Plugin extends AbsBaseAp
 
             /* Misc. cache behaviors. */
 
-            'allow_browser_cache' => '0', // `0|1`.
-            'when_logged_in'      => '0', // `0|1|postload`.
-            'get_requests'        => '0', // `0|1`.
-            'feeds_enable'        => '0', // `0|1`.
-            'cache_404_requests'  => '0', // `0|1`.
-            'cache_nonce_values'  => '0', // `0|1`.
+            'allow_browser_cache'               => '0', // `0|1`.
+            'when_logged_in'                    => '0', // `0|1|postload`.
+            'get_requests'                      => '0', // `0|1`.
+            'feeds_enable'                      => '0', // `0|1`.
+            'cache_404_requests'                => '0', // `0|1`.
+            'cache_nonce_values'                => '0', // `0|1`.
+            'cache_nonce_values_when_logged_in' => '0', // `0|1`.
 
             /* Related to exclusions. */
 
--- a/src/includes/closures/Ac/NcDebugUtils.php
+++ b/src/includes/closures/Ac/NcDebugUtils.php
@@ -137,11 +137,11 @@ $self->maybeGetNcDebugInfo = function ($reason_code = '', $reason = '') use ($se
                 break; // Break switch handler.
 
             case NC_DEBUG_IS_LOGGED_IN_USER_NONCE:
-                $reason = __('because the current page contains `_wpnonce` or `akismet_comment_nonce`. While your current configuration states that pages SHOULD be cache for logged-in visitors, `*nonce*` values in the markup (because these expire automatically) are not cache-compatible. See http://wsharks.com/1O1Kudy', SLUG_TD);
+                $reason = __('because the current page contains `_wpnonce` or `akismet_comment_nonce`. While your current configuration states that pages SHOULD be cache for logged-in visitors, `*nonce*` values in the markup are not cache-compatible. See http://wsharks.com/1O1Kudy for further details.', SLUG_TD);
                 break; // Break switch handler.
 
             case NC_DEBUG_PAGE_CONTAINS_NONCE:
-                $reason = __('because the current page contains `_wpnonce` or `akismet_comment_nonce`. Note that `*nonce*` values in the markup (because these expire automatically) are not cache-compatible. See http://wsharks.com/1O1Kudy', SLUG_TD);
+                $reason = __('because the current page contains `_wpnonce` or `akismet_comment_nonce`. Note that `*nonce*` values in the markup are not cache-compatible. See http://wsharks.com/1O1Kudy for further details.', SLUG_TD);
                 break; // Break switch handler.
 
             case NC_DEBUG_NO_USER_TOKEN:
--- a/src/includes/closures/Ac/NcDebugUtils.php
+++ b/src/includes/closures/Ac/NcDebugUtils.php
@@ -137,11 +137,11 @@ $self->maybeGetNcDebugInfo = function ($reason_code = '', $reason = '') use ($se
                 break; // Break switch handler.
 
             case NC_DEBUG_IS_LOGGED_IN_USER_NONCE:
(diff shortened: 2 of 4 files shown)


raamdev:
@jaswsinc Commit https://github.com/websharks/zencache-pro/commit/d9496fed1867f33604fde32a3c69b78686c77a29 does not include a UI to control this option, so as of now it can only be controlled via `wp-config.php`, e.g., 

```php
define('ZENCACHE_CACHE_NONCE_VALUES', TRUE);
```

However, my feeling is that we should add something to the Plugin Options UI that allows a site owner to easily control this setting. At the same time, a whole new "Nonce Caching" panel seems like overkill, and I also wouldn't want to put such an option inside "Enable/Disable" where it would be prominent. Maybe a new "Prevent Nonce Caching?" section could go inside "Theme/Plugin Developers"?

jaswsinc:
PR looks good to me. I think what you have is good enough. I don't see there being many legitimate use cases for this at all. So great to have it in there, but having a UI option for anything related to this feels like overkill to me also.

raamdev:
> I don't see there being many legitimate use cases for this at all.

Hmm, well given that as soon as we started excluding pages that contain `*nonce` in the content we started having users asking us how to turn that off, it seems like the number of cases is rather high (and we haven't even released this to 70,000+ ZenCache Lite users yet...).

Adding something to your `wp-config.php` file might be easy for most ZenCache Pro users, but it's _not_ easy for most ZenCache Lite users, hence the reason I feel we should have a way to disable this in the UI, even if it's buried a bit...

---

On this same topic: I just realized that if you simple _write_ about `_wpnonce` on your site, i.e., your post content contains that word, ZenCache picks that up and prevents caching, even if there is no real `nonce` being used...

jaswsinc:
> Adding something to your wp-config.php file might be easy for most ZenCache Pro users, but it's not easy for most ZenCache Lite users, hence the reason I feel we should have a way to disable this in the UI, even if it's buried a bit...

My feeling is that this really should _not_ be configurable at all, and if it is configurable in some way, it shouldn't be easy to do, or ever be suggested by us in any way. Caching nonce values is a bad idea.

> started having users asking us how to turn that off

Right. Well, it might come as a surprise to ZC users who have been seeing the cached copy of their pages with nonce values. However, the way it works now is _right_ way to handle this scenario.

If you disable that check and allow nonce values to be cached, you are eventually going to have bigger problems than a dynamic page not being cached, because if an nonce value is cached and it later expires, actions that you attempt to make while working on your site could start to fail.

So I'd rather see complaints about this, and find other ways to improve the experience, than to simply allow a site owner to disable the functionality; because that will only lead to other problems and/or mysterious bug reports that could actually be related to them having disabled nonce detection.

> On this same topic: I just realized that if you simply write about _wpnonce on your site, i.e., your post content contains that word, ZenCache picks that up and prevents caching, even if there is no real nonce being used...

I definitely agree there is room for some improvement there. If we can improve the way that we detect nonce values to eliminate some false-positives, that'd be great.

raamdev:
> if an nonce value is cached and it later expires, actions that you attempt to make while working on your site could start to fail.

If that were such a big issue, why have we not heard about this _at all_? We've only had one report, from someone specifically using bbPress + ZenCache + Logged-In User Caching.

On the other hand, we've had no fewer than half-a-dozen complains since v151220, asking us how to disable the `nonce` exclusion.

---

I understand what you're saying about excluding those pages being the _right_ thing to do, but my feeling is we can't just start excluding them without giving site owners a way of disabling that exclusion (along with being warned what that means). Once that new default option (excluding `nonce`) has been in the wild for a few months, we can get rid of the (yet-to-be-added) UI option and then allow people to disable the exclusion only via a filter.

jaswsinc:
> If that were such a big issue, why have we not heard about this at all?

Not getting reports about this doesn't surprise me, because it's not something that you can easily connect to ZenCache, or even reproduce consistently. It took us really digging in to even discover this ourselves, so I wouldn't expect to have seen a lot of reports about this particular problem.

However, once you realize what caching nonce values can lead to, it's very easy to reproduce it. I would argue that the number of reports we are seeing now (i.e., now that you can more easily diagnose it), is  proof of this having caused problems for people in the past. If you are navigating your site and nonce values are not getting cached, that means previously they _were_ being cached; and that's almost certainly a bad thing. Actions that depend on dynamic nonce values are not seeing what they expect to see, which can lead to all sorts of weird problems.

I also would not rule out the potential for this to cause a major security issue, should we allow nonce values to be cached. I'm not aware of it ever having been a major security issue, but it worries me to see a unique user-specific security token in any cache file, ever.

raamdev:
@jaswsinc writes...
> I also would not rule out the potential for this to cause a major security issue, should we allow nonce values to be cached. I'm not aware of it ever having been a major security issue, but it worries me to see a unique user-specific security token in any cache file, ever.

So I suppose the only reason we haven't seen WP Super Cache (maintained by Automattic itself) and W3TC address the `nonce` issue is that they don't support Logged-In User caching, is that correct?

If the `nonce` exclusion were limited to when Logged-In User caching was enabled, that would be one thing, but I think we're seeing a lot of people who are running plugins that add `nonce` values to the markup _even when the visitor is not logged-in_. (Is there ever a legitimate reason for doing that?)

---

I suppose at the very least what needs to happen is we need a KB Article that explains _why_ `nonce` values are excluded, and a link to that KB Article in the HTML notes that indicate caching has been disabled because of `nonce` in the markup.

raamdev:
@jaswsinc I was just working on a KB Article for _Why does ZenCache avoid caching nonce values?_ and I had a thought:

When Logged-In User caching is enabled, isn't there less of a security issue with caching nonce values? If each user has their own cache file then the nonce values that were generated for them are being stored in their own cache file. 

And if that's the case, then the only issue with caching nonce values for Logged-In Users is that they may experience issues with actions that contain expired nonce values, correct?

If we prevent nonce-caching for Logged-In Users (as we do now), then the Logged-In User caching feature of ZenCache becomes mostly useless, as many sites that enable Logged-In User caching are going to have caching disabled due to the presence of a nonce value somewhere in the markup.

---

What do you think about adding another `ZENCACHE_CACHE_LOGGED_IN_NONCE_VALUES` option to specifically enable/disable nonce caching for logged-in users? And what about adding a section to the **Logged-In Users** panel that allows site owners to allow caching of nonce values for logged-in users (with the default set to "No")? That section would include a warning about what that means, and a suggestion for lowering the cache expiration time to reduce chances of a cached nonce value expiring.

jaswsinc:
> And if that's the case, then the only issue with caching nonce values for Logged-In Users is that they may experience issues with actions that contain expired nonce values, correct?

Agreed.

> If we prevent nonce-caching for Logged-In Users (as we do now), then the Logged-In User caching feature of ZenCache becomes mostly useless, as many sites that enable Logged-In User caching are going to have caching disabled due to the presence of a nonce value somewhere in the markup.

Having an nonce in the markup on certain pages would be a strong indication of there being dynamic content and/or content associated with dynamic actions that can and probably will be taken on that page. So I don't see this as a big negative. An nonce is dynamic content, and also a subtle form of encryption too. It should not get cached IMO.

> What do you think about adding another ZENCACHE_CACHE_LOGGED_IN_NONCE_VALUES option to specifically enable/disable nonce caching for logged-in users? And what about adding a section to the Logged-In Users panel that allows site owners to allow caching of nonce values for logged-in users (with the default set to "No")? That section would include a warning about what that means, and a suggestion for lowering the cache expiration time to reduce chances of a cached nonce value expiring.

I have no objection to this. I can't say I'm a huge fan of the idea, but your logic sounds good to me. I'm not a fan, mostly because it introduces complexity and could lead to confusion. If it's possible to disable nonce detection through a filter, great. Introducing it as a part of the UI seems like overkill to me. Just my 2 cents. Totally your call. I see nothing wrong with the idea.

jaswsinc:
This portion of the article on WP nonce values might be worth considering. It looks like the default minimum age is around 12 hours. https://codex.wordpress.org/WordPress_Nonces#Modifying_the_nonce_lifetime

raamdev:
> It looks like the default minimum age is around 12 hours.

Actually, it's 1 day: https://github.com/WordPress/WordPress/blob/4.4/wp-includes/pluggable.php#L1789-L1812

raamdev:
> minimum age

Oh, sorry, I missed that part. :laughing:

PR #204, opened 2015-12-28
Only add/remove from htaccess file when necessary

raamdev:
See websharks/zencache#641

---
_Quoting @raamdev from https://github.com/websharks/zencache/issues/641#issuecomment-167659466:_

> After some research, I discovered a few issues with the htaccess utilities in ZenCache Pro v151220:
> 
> - Every time the plugin options are saved when ZenCache is enabled, ZenCache [calls `addWpHtaccess()`](https://github.com/websharks/zencache-pro/blob/151220/src/includes/classes/Actions.php#L570-L572), regardless of whether or not it actually needs to (i.e., whether or not any options are enabled that require adding something to the htaccess file).
> - ZenCache does not intelligently keep track of which options are currently enabled that require htaccess rules, and as a result [more code than necessary](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L21-L42) is being run each time the site owner saves the plugin options and `addWpHtaccess()` is called.
> 
> As a result of all this unnecessary work, I believe we're increasing the chances that the site owner might see an error message when the error message isn't even applicable (i.e., ZenCache has no reason to add/remove anything from the htaccess file).
> 
> I've submitted a PR (https://github.com/websharks/zencache-pro/issues/204) that resolves these issues.

### DIFF
--- a/src/includes/closures/Plugin/HtaccessUtils.php
+++ b/src/includes/closures/Plugin/HtaccessUtils.php
@@ -24,7 +24,7 @@ $self->addWpHtaccess = function () use ($self) {
         return false; // Not running the Apache web server.
     }
     if (!$self->options['enable']) {
-        return false; // Nothing to do.
+        return true; // Nothing to do.
     }
     if (!$self->removeWpHtaccess()) {
         return false; // Unable to remove.
--- a/src/includes/closures/Plugin/HtaccessUtils.php
+++ b/src/includes/closures/Plugin/HtaccessUtils.php
@@ -16,6 +16,8 @@ $self->htaccess_marker = 'WmVuQ2FjaGU';
 * @since 15xxxx Improving `.htaccess` tweaks.
 *
 * @return array Plugin options that have associated htaccess rules
+*
+* @note We keep track of this to avoid the issue described here: http://git.io/vEFIH
 */
 $self->options_with_htaccess_rules = array('cdn_enable');
 
@@ -25,6 +27,8 @@ $self->options_with_htaccess_rules = array('cdn_enable');
  * @since 151114 Adding `.htaccess` tweaks.
  *
  * @return boolean True if added successfully.
+ *
+ * @TODO Improve error reporting detail to better catch unexpected failures; see http://git.io/vEFLT
  */
 $self->addWpHtaccess = function () use ($self) {
     global $is_apache;
@@ -86,6 +90,8 @@ $self->addWpHtaccess = function () use ($self) {
  * @since 151114 Adding `.htaccess` tweaks.
  *
  * @return boolean True if removed successfully.
+ *
+ * @TODO Improve error reporting detail to better catch unexpected failures; see http://git.io/vEFLT
  */
 $self->removeWpHtaccess = function () use ($self) {
     global $is_apache;
@@ -148,25 +154,6 @@ $self->needHtaccessRules = function () use ($self) {
         }
     }
     return false; // No, there are no options enabled that require htaccess rules.
-};
-
-/*
- * Utility method used to unlock and close htaccess file resource.
- *
- * @since 151114 Adding `.htaccess` tweaks.
- *
- * @param array $htaccess                   Array containing at least an `fp` file resource pointing to htaccess file.
- *
- * @return bool False on failure, true otherwise.
- */
-$self->closeHtaccessFile = function (array $htaccess) use ($self) {
-    if (!is_resource($htaccess['fp'])) {
-        return false; // Failure; requires a valid file resource.
-    }
-    flock($htaccess['fp'], LOCK_UN);
-    fclose($htaccess['fp']);
-
-    return true;
 };
 
 /*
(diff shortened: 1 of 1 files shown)


raamdev:
@jaswsinc A review here when you get a chance. :smile: Thanks!

jaswsinc:
@raamdev I reviewed the code for these changes.

On [this line](https://github.com/websharks/zencache-pro/blob/fa7bbfc38198b6814cd59341cc0ad7646174b925/src/includes/closures/Plugin/HtaccessUtils.php#L39) I think the call to `findHtaccessMarker()` is redundant, since that is the responsibility of `removeWpHtaccess()` already.

I also think that [this block](https://github.com/websharks/zencache-pro/blob/fa7bbfc38198b6814cd59341cc0ad7646174b925/src/includes/closures/Plugin/HtaccessUtils.php#L44-L46) should always be run first, before we check if there is the need to add anything; i.e., remove what was there before we consider adding anything new.

Otherwise it looks good to me.

---

## Other Thoughts

I'm a fan of optimization, and I like where you are going with this. However, in this particular case my feeling is that `needHtaccessRules()` is unnecessary. The `addWpHtaccess()` method already has this job to do; i.e., as it cycles through each of the template blocks to see if they apply.

So while I don't see anything wrong with `needHtaccessRules()`, my feeling is that it will add more complexity than is necessary. There are likely to be more and more use cases for minor `.htaccess` tweaks as we move forward, so having two different routines to work out the logic is going to make it overly complex in my view.

In [this line](https://github.com/websharks/zencache-pro/blob/fa7bbfc38198b6814cd59341cc0ad7646174b925/src/includes/closures/Plugin/HtaccessUtils.php#L67-L70) we already bail if nothing needs to be written.

Also, this only runs whenever you update your ZC options, right? So it's not an area of the codebase where optimization is really important.

raamdev:
> On this line I think the call to findHtaccessMarker() is redundant, since that is the responsibility of removeWpHtaccess() already.
> 
> I also think that this block should always be run first, before we check if there is the need to add anything; i.e., remove what was there before we consider adding anything new.

I disagree, and here's why:

As of now, if a site owner does not have CDN Filters enabled there is _no reason whatsoever_ to read or write to the `.htaccess` file. 

And yet, as of now, here's what we do _**every time the site owner saves the plugin options even when there's nothing for us to do with the `.htaccess` file**_:

- [Call `removeWpHtaccess()`](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L29), which then [reads the `.htaccess` file to look for the unique marker](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L84), and then [reads the `.htaccess` file and acquires an exclusive lock](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L87), and then [searches the `.htaccess` file for our comment markers and removes them if found and then writes to the `.htaccess` file and releases the lock](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L91-L94).
- [Call `readHtaccessFile()`](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L32), which then [reads the `.htaccess` file and acquires an exclusive lock](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L87), and then assuming nothing needs to be added to the `.htaccess` file, [calls `closeHtaccessFile()` which releases the lock](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/HtaccessUtils.php#L53).

That means we read the `.htaccess` file three (3) times, acquire two (2) separate exclusive locks, and write to the `.htaccess` file once (1). All of that work _when we can know ahead of time that we don't need to touch the `.htaccess` file_. 

That was my reason for adding `needHtaccessRules()`.

I have seen a handful of reports (including one from Pat) where there was an error message displayed on the Dashboard related to being unable to read/write to the `.htaccess` file, in a scenario where there was no reason to do so in the first place (i.e., CDN Filters were not enabled). My feeling is that because of all the extra unnecessary work we're doing each time the site owner saves the plugin options, we're occasionally tripping things up and something along the line fails and produces the error message (which goes away on the next refresh because it works fine the next time you try it).

jaswsinc:
> All of that work when we can know ahead of time that we don't need to touch the .htaccess file.

That's a good point and I agree.

> I have seen a handful of reports (including one from Pat) where there was an error message displayed on the Dashboard related to being unable to read/write to the .htaccess file

At some point it might be a good idea to add some more ways for this to wrong; i.e., increase the number of detailed exceptions that we throw until we can pinpoint more specifically what the problem is.

raamdev:
@jaswsinc I added a few TODOs in https://github.com/websharks/zencache-pro/commit/f887948830c91b06705cb973ff625860672682f1 so that we can revisit improving error reporting detail in the Htaccess Utils; I agree we need to do better there.

PR #205, opened 2015-12-28
Show Dashboard notice when ACE is enabled and PHP allow_url_fopen=0

raamdev:
See websharks/zencache#644

---

**Note:** This PR also prevents the Auto-Cache Engine from running when `allow_url_fopen=0`; see https://github.com/websharks/zencache-pro/commit/461435154569927b1b78de9a717b67c4c9bc4715


raamdev:
@jaswsinc A quick review here, please. :grin:

jaswsinc:
@raamdev This PR looks great to me as-is. Nice!

While reviewing this PR, I noticed something else that could use improvement.

After [this line](https://github.com/websharks/zencache-pro/blob/151220/src/includes/closures/Plugin/NoticeUtils.php#L292) please add the following:

  ```php
  ksort($notice); // For more accurate comparison in other routines.
  ```

raamdev:
> After this line please add the following:

@jaswsinc Done. Thanks for the review! :-)

PR #208, opened 2016-01-22
Feature Request: Option to disable HTML Compression for Logged-In Users

renzms:
see: websharks/zencache#650


src/includes/classes/Plugin.php
```diff
@@ -345,6 +346,7 @@ public function setup()
             'htmlc_compress_css_code'              => '1', // `0|1`.
             'htmlc_compress_js_code'               => '1', // `0|1`.
             'htmlc_compress_html_code'             => '1', // `0|1`.
+            'htmlc_when_logged_in' => '0', // `0|1`; enable when logged in?
```
raamdev: @renzms I noticed the formatting is slightly off here; notice how your addition here doesn't line up with the options above it.

raamdev:
@renzms It looks like all that's needed now is to check this option before running the HTML Compressor when users are logged in. Can you find where that would go?

renzms:
@raamdev 
> I noticed the formatting is slightly off here; notice how your addition here doesn't line up with the options above it.
Thanks, I didn't catch the alignment in Atom.

>It looks like all that's needed now is to check this option before running the HTML Compressor when users are logged in. Can you find where that would go?

Where would that go?

raamdev:
> Where would that go?

Were you able to look? Any ideas where it might go?

renzms:
> Were you able to look? Any ideas where it might go?

After this: [here](https://github.com/websharks/zencache-pro/blob/160103/src/includes/templates/advanced-cache.txt#L407)?

src/includes/classes/MenuPageOptions.php
```diff
@@ -840,6 +840,12 @@ public function __construct()
...
             echo '         <p>'.sprintf(__('<strong>Note:</strong> This does NOT impact the overall cache expiration time that you configure with %1$s. It only impacts the sub-routines provided by the HTML Compressor. In fact, this expiration time is mostly irrelevant. The HTML Compressor uses an internal checksum, and it also checks <code>filemtime()</code> before using an existing cache file. The HTML Compressor class also handles the automatic cleanup of your cache directories to keep it from growing too large over time. Therefore, unless you have VERY little disk space there is no reason to set this to a lower value (even if your site changes dynamically quite often). If anything, you might like to increase this value which could help to further reduce server load. You can <a href="https://github.com/websharks/HTML-Compressor" target="_blank">learn more here</a>. We recommend setting this value to at least double that of your overall %1$s expiration time.', SLUG_TD), esc_html(NAME)).'</p>'."\n";
+            echo '      <h3>'.__('Do NOT use Html Compression for logged in users?', SLUG_TD).'</h3>'."\n";
```
raamdev: @renzms Please change this to "Do NOT use HTML Compression for logged-in users?" (Note the changes to: 'HTML' and 'logged-in'.)

raamdev: @renzms Actually, please change it to the following (I just realized we're going to be defaulting this option to disabled, so the title should reflect that); change it to the following please:

> Enable HTML Compression for logged-in users?

src/includes/classes/MenuPageOptions.php
```diff
@@ -840,6 +840,12 @@ public function __construct()
...
+            echo '      <p>'.__('HTML compression allows for cached pages to load a lot faster compared to the initial visit to the page. Logged in users may have a much slower first time visit of loading of a page because of the HTML Compression. Compared to regular visitors or users who are not logged in, those visitors/users may have a faster experience because of the Auto-Cache Engine and the way it can pre-cache. In short, do NOT turn this off unless you know what you\'re doing.', SLUG_TD).'</p>'."\n";
```
raamdev: @renzms Please change this to the following:

> HTML compression allows for faster loading of cached pages however this should remain disabled for logged-in users because their user-specific cache has a much shorter Time To Live (TTL). For example, if you're logged into the site as a user and you submit a form, that triggers a clearing of the cache for that user (including the HTML Compressor cache). Lots of little actions you take can result in a clearing of the cache. This shorter TTL is not ideal when running the HTML Compressor. The HTML Compressor does a deep analysis of the page content and the associated resources in order to compress things for you intelligently. For logged-in users, it is better to skip that and just cache the HTML source as-is, avoiding that extra overhead. In short, do NOT turn this on unless you know what you\'re doing.

renzms: > @renzms Please change this to the following:
> 
> HTML compression allows for faster loading of cached pages however this should remain disabled for logged-in users because their user-specific cache has a much shorter Time To Live (TTL). For example, if you're logged into the site as a user and you submit a form, that triggers a clearing of the cache for that user (including the HTML Compressor cache).

@raamdev Updated and changed the line to:  "For example, if you\'re logged into the site as a user and you submit a form, that triggers a clearing of the cache for that user (including the HTML Compressor cache)." so it would parse correctly.

raamdev:
> After this: line here?

Right, except I would put a block [just like this one](https://github.com/websharks/comet-cache-pro/blob/160223.1/src/includes/templates/advanced-cache.txt#L240-L251) right underneath it for `HTMLC_WHEN_LOGGED_IN` and then there should be a conditional just like [this one](https://github.com/websharks/comet-cache-pro/blob/160223.1/src/includes/closures/Ac/HtmlCUtils.php#L20-L22) right underneath it that checks `if ($self->is_user_logged_in && !COMET_CACHE_HTMLC_WHEN_LOGGED_IN)`.

renzms:
> Right, except I would put a block just like this one right underneath it for HTMLC_WHEN_LOGGED_IN and then there should be a conditional just like this one right underneath it that checks if ($self->is_user_logged_in && !COMET_CACHE_HTMLC_WHEN_LOGGED_IN).

@raamdev Added changes, ready for review! Thanks.

src/includes/classes/Plugin.php
raamdev: @renzms I noticed the formatting is still off on this line.

raamdev:
@renzms I added another note here: https://github.com/websharks/comet-cache-pro/pull/208#discussion_r54090185

renzms:
@raamdev Spacing fixed, ready for review! (again :+1: ) - https://github.com/websharks/comet-cache-pro/commit/142c3cee5c0bf24765b365cd5c62143e53935873

raamdev:
@renzms I improved the options panel a bit; see https://github.com/websharks/comet-cache-pro/commit/acae0865bd046b6dbde432d0a3c303f1ed4f6f60. 

Here's a screenshot of the new section in **HTML Compression** as of now:

raamdev:
@jaswsinc This is ready for a 2nd reviewer when you get a chance. :-) I've tested this feature branch to confirm it works as expected.

raamdev:
Updating screenshot; fixed a few things in the description.

raamdev:
@jaswsinc Final screenshot, sorry:

jaswsinc:
@raamdev I reviewed your UI enhancements with the improved description of this feature. Looks great to me. Nice!

PR #211, opened 2016-01-30
Rebrand as IntelliCache

raamdev:


lkraav:
Are you sure about the name?

http://blog.citrix24.com/how-does-intellicache-work/ seems like you could be tangling with some big boys.

raamdev:
@lkraav We do know about the other usage, but it's just being used as a feature name not a product name. We're being forced to change our name (we'll be publishing an announcement soon) and IntelliCache is what we've decided to go with.

jaswsinc:
:+1: Love the new name.

jaswsinc:
_Working on the second part of this now._

raamdev:
@jaswsinc Back ZenCache compat. ready for review; see https://github.com/websharks/zencache-pro/compare/66a2dfcd3c46...dca6df1e3229

raamdev:
@jaswsinc Thanks for the review. I applied your suggestions here: https://github.com/websharks/zencache-pro/commit/264dd4b86b343d45d60deddf4337b236a8e2c309

jaswsinc:
Those looks great.

I noticed [here](https://github.com/websharks/zencache-pro/blob/intellicache/src/includes/classes/VsUpgrades.php#L181-L192) that these option key checks are no longer necessary. I think that routine is based on the transition from Quick Cache to ZenCache, and those option keys where different during that switch. Now they are identical. So those lines can come out now.

jaswsinc:
There are a couple of comments in the `closures/shared/HookUtils.php` file that contain `Quick Cache`.

jaswsinc:
After [this line](https://github.com/websharks/zencache-pro/blob/intellicache/src/includes/classes/Conflicts.php#L35) add:

```php
'zencache', 'zencache-pro',
```

and also add those two slugs to the array [here](https://github.com/websharks/zencache-pro/blob/intellicache/src/includes/classes/Conflicts.php#L48).

raamdev:
@jaswsinc Done. https://github.com/websharks/zencache-pro/commit/a1e331819da38714141b1553d96ba666c7d9a6e2

jaswsinc:
That looks ready to merge now. Those last changes look great, despite me linking you to the wrong line by mistake. Cool :-)

jaswsinc:
Oh. Nevermind. I thought I linked you to the wrong line, but I didn't. There is a new line there now after your most recent changes.

:laughing: My only mistake was thinking I made a mistake.

PR #213, opened 2016-02-09
Rebrand as Comet Cache

raamdev:
## TODO 

- [x] `si-zencache-logo` and `si-zencache` in `src/client-s/css/menu-pages.min.css` need to be updated to point to a Comet Cache graphic, which requires that they be added to the Sharkicons library.
- [x] Ensure that the `.htaccess` routines that look for the `ZenCache` marker in the `.htaccess` file can properly handle a transition to Comet Cache.
- [x] Fix issue with build script not generating the translation file properly; see https://github.com/websharks/zencache-pro/pull/213#issuecomment-181994597; **DONE:** the repo directory must match the plugin PHP file name (`comet-cache-pro`) for the `wp-i18-tools` to properly generate the POT file.
- [x] Fix issue with Admin Bar menu items ('Stats' and 'Clear Cache') not working.


raamdev:
@jaswsinc This is ready for another set of eyes.

---

One odd thing that's happening is with the translation file when I build the package from this branch:

Here's what should get generated at the top:

```
# Copyright (C) 2016 ZenCache Pro
# This file is distributed under the same license as the ZenCache Pro package.
msgid ""
msgstr ""
"Project-Id-Version: ZenCache Pro 160103\n"
"Report-Msgid-Bugs-To: http://wordpress.org/tag/zencache-pro\n"
"POT-Creation-Date: 2016-01-03 20:06:07+00:00\n"
```

Here's what actually gets generated:

```
# Copyright (C) 2016 
# This file is distributed under the same license as the  package.
msgid ""
msgstr ""
"Project-Id-Version:  PHP 5.4</a>', 'comet-cache').'</p>';\n"
"Report-Msgid-Bugs-To: http://wordpress.org/tag/zencache-pro\n"
"POT-Creation-Date: 2016-02-09 16:48:07+00:00\n"
```

Notice the `Project-Id-Version:`; it doesn't look right. I double-checked the updated `.build.props` file and it looks right. Still digging into this; not sure what's going on but maybe something will catch your eye. :-)

raamdev:
@jaswsinc Note that I'm maintaining a list of TODOs over here: https://github.com/websharks/cometcache.com/issues/86

raamdev:
@jaswsinc Regarding this point:

> - Ensure that the `.htaccess` routines that look for the `ZenCache` marker in the `.htaccess` file can properly handle a transition to Comet Cache.

I'm thinking what we need is just an upgrade routine in Comet Comet cache that considers an existing ZenCache block in the `.htaccess` and then either removes that, or updates the `ZenCache` to `Comet Cache` so that the htaccess utils work as expected. Does that sound right?

jaswsinc:
> Notice the Project-Id-Version:; it doesn't look right. I double-checked the updated .build.props file and it looks right. Still digging into this; not sure what's going on but maybe something will catch your eye. :-)

The `wp-i18n-tools` package pulls the plugin name and some other details from the plugin PHP file. If you ran this from a `zencache-pro` folder, it was probably looking for a plugin file named `zencache-pro.php` and could not find it. That might be why it is missing the name of the plugin.

jaswsinc:
> I'm thinking what we need is just an upgrade routine in Comet Comet cache that considers an existing ZenCache block in the .htaccess and then either removes that, or updates the ZenCache to Comet Cache so that the htaccess utils work as expected. Does that sound right?

Yes, that sounds right to me.

raamdev:
> If you ran this from a zencache-pro folder, it was probably looking for a plugin file named zencache-pro.php and could not find it. That might be why it is missing the name of the plugin.

Confirmed. That was the problem. I temporarily renamed the repo directory from `zencache-pro` to `comet-cache-pro` and re-ran the build. The `wp-i18-tools` generated the POT file correctly.

raamdev:
@jaswsinc This is ready for some more testing, specifically the Lite version. (Note that the Lite version does not write anything to the `.htaccess` file, and never has, so there's no need to ensure upgrade routines related to `.htaccess` files are in place yet.)

jaswsinc:
> This is ready for some more testing, specifically the Lite version. (Note that the Lite version does not write anything to the .htaccess file, and never has, so there's no need to ensure upgrade routines related to .htaccess files are in place yet.)

I just built and run the lite version of Comet Cache through some basic tests and reviewed the Dashboard. It looks and works great!

jaswsinc:
@raamdev I noticed, while working on Comment Mail, that we should add `comet-cache` and `comet-cache-pro` to this array of slugs which are OK to deactivate automatically should that create a conflict. Referencing: https://github.com/websharks/zencache-pro/blob/cc/src/includes/classes/Conflicts.php#L48

jaswsinc:
@raamdev I noticed that you chose to use an `_` in the `Comet_Cache` sub-namespace name.

I was just looking over the PSR-4 and PSR-2 style guidelines to be sure this won't throw any warnings. It looks like it's fine. I also ran a couple of files through PHP Code Sniffer to see if it triggered any warnings about this, and that looks fine too.

The reason the `_` character threw me at first, is because the `_` is used in PSR-4 autoloaders for backward compatibility with directory structures that are difficult to work with; e.g., Composer will convert `_` into `/` whenever it looks for class files. However, that autoload conversion is limited to the class name itself, so `Comet_Cache` in the sub-namespace is still OK, so long as we don't use `_` in any of the base class names.

jaswsinc:
@raamdev We should remove [this line](https://github.com/websharks/zencache-pro/blob/cc/src/includes/stub.php#L30) in Comet Cache. I noticed, while working on Comment Mail, that this line can cause a minor problem whenever both the lite and pro versions are running at the same time. Our conflicting plugin checks handle this gracefully at the moment, but I suggest removing it anyway.

raamdev:
@jaswsinc writes...
> I noticed that you chose to use an `_` in the `Comet_Cache` sub-namespace name.

I'm fine with changing it to `WebSharks\\CometCache` if you think that's better. I only went with `Comet_Cache` to maintain consistency with what `GLOBAL_NS` is set to (`comet_cache`), which _should_ use an underscore because it's used in WP option key names. 

> so long as we don't use `_` in any of the base class names.

Can you give me an example of where we use the plugin name in a base class name anyway? With everything namespaced, we don't need to use the plugin name. The only place I know of that we use the plugin name is in the API, in `src/includes/classes/ApiBase.php`, e.g., `comet_cache:clear()`.

That said, if we did change the sub-namespace to use `CometCache`, that might be a subtle reminder if/when we create a new class that the name should not use the underscore.

raamdev:
> we should add comet-cache and comet-cache-pro to this array of slugs

Ah, good idea. Done. https://github.com/websharks/zencache-pro/commit/4a3c53cc148c3567ceb2b9da22acfd59d9e8593f

raamdev:
> We should remove this line in Comet Cache. I noticed, while working on Comment Mail, that this line can cause a minor problem whenever both the lite and pro versions are running at the same time. Our conflicting plugin checks handle this gracefully at the moment, but I suggest removing it anyway.

Done. https://github.com/websharks/zencache-pro/commit/1bffb2f33324132e2d7289bc306a193070d77693

jaswsinc:
> Can you give me an example of where we use the plugin name in a base class name anyway? With everything namespaced, we don't need to use the plugin name. The only place I know of that we use the plugin name is in the API, in src/includes/classes/ApiBase.php, e.g., comet_cache:clear().

Yep. Shouldn't be needed. Like you said, we have the sub-namespace name for this, and since Composer doesn't convert `_` into `/` in the sub-namespace name, it's going to be just fine. Which is cool, because I didn't even realize that was possible in Composer until now.

However, in the PSR guidelines we have `snake_case` for functions, `camelCase` for method names, and `StudlyCaps` for classes. For that reason, I think it's probably better if we don't mix StudlyCaps with Studly_Snake_Case; even though I now see that it's perfectly legal to do that.

> That said, if we did change the sub-namespace to use CometCache, that might be a subtle reminder if/when we create a new class that the name should not use the underscore.

Yeah. Since classes, traits, and interfaces must use StudlyCaps, I think it's more consistent if we make the entire class path StudlyCaps also. I have observed that most PHP projects following PSR do the same, so we should be in good company there; i.e., less likely to conflict with something in the future if we leave the `_` character out and just let StudlyCaps be the separator.

The reason for `GLOBAL_NS` being set to `comet_cache`, is because that's intended primarily as the Global Variable namespace. Variables in all of source code use `snake_case`; i.e., `$comet_cache` is what `GLOBAL_NS` results in; e.g., `$GLOBALS['comet_cache']` or `global $comet_cache;`

jaswsinc:
There is one exception to `$snake_case` in variable names. It has nothing to do with the PSR guidelines though. PSR doesn't have a standard for variable names. Most people use `$snake_case`.

Here's what I started doing in websharks/core, websharks/wp-core, and in websharks/s2member-x.

In those projects, I'm using `$StudlyCaps` for variables that represent class object instances, because it makes it easier to read the source code and know that a variable is a reference to an object class instance that way, which also uses StudlyCaps. I haven't done this in any of our other projects though. Just seems not worth the trouble.

Example here: https://github.com/websharks/core/blob/000000-dev/src/includes/classes/Utils/HashIds.php#L48

raamdev:
@jaswsinc writes...
> However, in the PSR guidelines we have snake_case for functions, camelCase for method names, and StudlyCaps for classes. For that reason, I think it's probably better if we don't mix StudlyCaps with Studly_Snake_Case; even though I now see that it's perfectly legal to do that.

Agreed. I've changed it here: https://github.com/websharks/zencache-pro/commit/a9759b2aa779f54ae5748255fae4f06e057a3047

raamdev:
@jaswsinc While testing ZenCache Lite → Comet Cache Lite migration, I discovered that a fatal error is thrown:

```
[19-Feb-2016 19:35:09 UTC] PHP Fatal error:  Cannot redeclare class zencache in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/zencache/includes/api-class.php on line 19
[19-Feb-2016 19:35:09 UTC] PHP Stack trace:
[19-Feb-2016 19:35:09 UTC] PHP   1. {main}() /Users/raam/Sites/zcpro.dev/wordpress/wp-admin/plugins.php:0
[19-Feb-2016 19:35:09 UTC] PHP   2. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-admin/plugins.php:10
[19-Feb-2016 19:35:09 UTC] PHP   3. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-admin/admin.php:31
[19-Feb-2016 19:35:09 UTC] PHP   4. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-load.php:37
[19-Feb-2016 19:35:09 UTC] PHP   5. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-config.php:88
[19-Feb-2016 19:35:09 UTC] PHP   6. include_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-settings.php:254
[19-Feb-2016 19:35:09 UTC] PHP   7. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/zencache/zencache.php:74
[19-Feb-2016 19:35:09 UTC] PHP   8. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/zencache/zencache.inc.php:2481
```

This is because [`includes/api-class.php`](https://github.com/websharks/zencache/blob/151107/zencache/includes/api-class.php) is attempting to redeclare the `zencache` class _after_ Comet Cache checks if the `zencache` class exists and then [creates a `zencache` class alias](https://github.com/websharks/zencache-pro/blob/cc/src/includes/api.php#L14-L16).

The Comet Cache API (and the conflicting `zencache` class alias) is [only loaded _after_](https://github.com/websharks/zencache-pro/blob/cc/src/includes/plugin.php#L14-L17) passing a conflicts check, which should be checking if ZenCache is active and deactivating it.

However, I think the problem is that the plugins are not deactivated immediately when the conflicts check is run, but rather [the deactivation is hooked onto `admin_init`](https://github.com/websharks/zencache-pro/blob/cc/src/includes/classes/Conflicts.php#L49-L53), which I'm running results in ZenCache Lite being deactivated too late, i.e., the PHP Fatal Error occurs first.

If that's the case, it would seem to me that we need to deactivate ZenCache Lite immediately, right when we do the conflict check, as opposed to hooking into `admin_init` with an anonymous function.

I just tested that (getting rid of the `add_action()` and just calling `deactivate_plugins()` immediately) and that seems to fix the problem. Can think of any reason not to do it that way?

raamdev:
@jaswsinc See https://github.com/websharks/zencache-pro/commit/944a535dabdbc16ad8c67aa184681dc97aa2f20a.

jaswsinc:
The only reason I was using `admin_init` there is because the `deactivate_plugins()` function is only available in the administrative area of WordPress; i.e., it is not available on the front-end. So what you have seems fine.

However, I think it would be even better if you move the `function_exists` check into the same line as the `in_array` check here: https://github.com/websharks/zencache-pro/commit/944a535dabdbc16ad8c67aa184681dc97aa2f20a#diff-44226fe2329a01455f5421b20930f970R48

That way if it is not possible to deactivate it immediately, we fail and list it is a conflict in the else condition that follows that section.

---

> I just tested that (getting rid of the add_action() and just calling deactivate_plugins() immediately) and that seems to fix the problem. Can you think of any reason not to do it that way?

I haven't tested that myself yet, but I'm wondering how that resolved the problem exactly. The reason it seems strange that this would resolve the problem, is because plugins are loaded with a foreach loop here: https://github.com/WordPress/WordPress/blob/4.4.2/wp-settings.php#L254

So deactivating the plugin immediately still won't stop that foreach loop at runtime.

raamdev:
Referencing discussion on Slack regarding the PHP Fatal error occurring when migrating from ZenCache Lite: https://websharks.slack.com/archives/comet-cache/p1455985897000368

raamdev:
PHP Fatal error has been tested and confirmed fixed. See https://github.com/websharks/zencache-pro/commit/c8c60c52ba4923856b6c047c9e0d03cabd9d7d65

PR #217, opened 2016-02-24
Fix bug in Auto-Cache Engine that prevented it from running properly.

raamdev:
See websharks/comet-cache#679


raamdev:
@jaswsinc Quick review here please, when you get a moment.

jaswsinc:
Looks good. Nice find! I read your explanation of the problem and I see that we need to match up with the `isAutoCacheEngine()` method where we are using `GLOBAL_NS`. Tricky!

PR #218, opened 2016-02-24
Fix bug in backwards compatibility for ZENCACHE_ALLOWED

raamdev:
See websharks/comet-cache#683


raamdev:
@jaswsinc Quick review when you get a chance please.

jaswsinc:
Looks great! Nice catch on the lowercase issue.

**One suggestion:** Since this runs on every page view, I would try to optimize those calls to `strtoupper()`. It seems easy enough to run that once at the top of this method and then reuse a variable that contains the uppercase version in the rest of the sub-routines.

raamdev:
> Since this runs on every page view, I would try to optimize those calls to strtoupper().

Ah, good idea. Thanks. Done: https://github.com/websharks/comet-cache-pro/commit/367206804724df6a6563ee3996cca685fc56e8f9

jaswsinc:
This looks great to me.

PR #219, opened 2016-02-26
Add backwards compatibility for $GLOBALS['zencache']

raamdev:
See websharks/comet-cache#689


raamdev:
@jaswsinc Quick review, please. I've tested this and it behaves as expected. Is there anything else you can think of that we might need to do here?

jaswsinc:
Looks good. Nothing else comes to mind, no.

PR #220, opened 2016-02-26
Improve support for Accelerated Mobile Pages via the AMP plugin

raamdev:
See websharks/comet-cache#688


jaswsinc:
This looks great to me. It might be worth adding an `@TODO` marker in this area of the codebase so we can remember that at some point it would be a good idea to try and detect the existence of the AMP plugin before we add this in. That would be a subtle enhancement that might help in a few edge cases in the future; e.g., maybe a new conditional method at some point, like `isAmpInstalled()`.

raamdev:
@jaswsinc Great idea. Done. Thanks!

PR #222, opened 2016-02-28
Convert to PHP Traits

raamdev:
See websharks/comet-cache#635

---

## Converting Closures to Traits (Step 1 of 2)

In this step we will convert Closure files into Traits. The codebase will _not_ function with these changes alone. In step 2 of 2 (later) we will implement the new Traits and be ready then. I will outline that soon.

### Outline for Step 1

- [x] Rename [this directory](https://github.com/websharks/comet-cache-pro/tree/000000-dev/src/includes/closures) to `traits`.

- [x] Convert each file in the `traits` directory into a Trait.

  - [Example file](https://github.com/websharks/comet-cache-pro/blob/000000-dev/src/includes/closures/Plugin/WcpUtils.php) converted to a Trait. Converted (in part, as an example): https://gist.github.com/jaswsinc/a175d369dfb0f17c876a

#### Converting Closures into Traits

Following the example above, here are some things to be aware of.

- [x] `$self->property = value;` becomes `public $property = value;`
- [x] `$self->functionName` becomes `public function functionName`
- [x] Ditch `use($self)`.
- [x] Change `$self` references inside a method into `$this`.
- [x] Change `};` ending a Closure into just `}`.
- [x] Wrap all of the properties/methods into a Trait with the same name as the file they live in; i.e., `trait WcpUtils {}` as seen in the example above.
- [x] Change the namespace in the file by adding `\Traits\Ac|Plugin|Shared` (depending on the main folder).

## Converting Closures into Traits (Step 1.3 of 2, lol)

- [x] Alter the namespace in these files by appending `\Classes` to the existing namespace. https://github.com/websharks/comet-cache-pro/tree/000000-dev/src/includes/classes

- [x] In `composer.json`, replace [this line](https://github.com/websharks/comet-cache-pro/blob/160227/composer.json#L46) with the following:

  ```json
  "WebSharks\\CometCache\\Pro\\Classes\\": "src/includes/classes/"
  "WebSharks\\CometCache\\Pro\\Traits\\": "src/includes/traits/"
  ```

- [x] At the top of each Trait file, add a blank line, and then the following after the namespace.

  ```php
  use WebSharks\CometCache\Pro\Classes;
  ```

- [x] In each Trait file, search for instances of `new ClassName` and prepend those with `Classes\`; i.e., `new Classes\ClassName` so that it can be resolved properly from inside the Trait.

- [x] Do the same thing for static calls; i.e., look for instances of `ClassName::` and change that to `Classes\ClassName::` instead, so that it will resolve properly from inside the Trait.


raamdev:
@jaswsinc A couple of questions:

---

- [ ] In a couple files (such as `traits/Plugin/UserUtils.php`) we have the following, which does not seem right:

```
$this->currentUserCanWipeCache = $this->currentUserCanClearCache;
```

---

- [ ] In `traits/Shared/ConditionalUtils.php`, we had 

```
return $self instanceof Plugin;
```

That got changed to `return $this instanceof Plugin;`; is that right?

---

- [ ] In `traits/Shared/HttpUtils.php` we have 

```
$this->{'http_status'}
```

Is that correct? Or should it be `$this->http_status`?

jaswsinc:
> $this->currentUserCanWipeCache = $this->currentUserCanClearCache;

You're right, that is incorrect. Those are aliases. Now that we are converting to traits, we need to change the way the alias works. Instead of simply pointing one property to another (i.e., a Closure), we need to create an alias method like this.

```php
public function currentUserCanWipeCache()
{
    return $this->currentUserCanClearCache();
}
```

In some cases, you will need to alias the arguments as well.

```php
public function aliasMethod($arg1, $arg2)
{
    return $this->realMethod($arg1, $arg2);
}
```

jaswsinc:
> That got changed to return $this instanceof Plugin;; is that right?

Yes, that is correct.

jaswsinc:
> $this->{'http_status'}
> Is that correct? Or should it be $this->http_status?

Either of those are correct; i.e., they both reference a property in `$this` class. However, I agree with you, it should be `$this->http_status` for clarity. I'm not sure why I have `{}` brackets there. The only time those are really needed is when a property breaks the rules in some way; e.g., if the property starts with a number or with another character that is not valid variable syntax, strictly speaking.

Another example of when `{}` brackets are helpful, is when you are building the property name out of another variable; e.g., `$this->{$dynamic.'_property_name'} = 'hello world';`. That may have been the case in this part of the code in the past. So this could be a remnant from something in the past that required the `{}` brackets. I don't see any reason to keep them moving forward though.

raamdev:
@jaswsinc another question:

- [ ]  `traits/Ac/NcDebugConsts.php` and `traits/Shared/CachePathConsts.php` cannot be wrapped in `trait {}`, as apparently it's not possible for constants to be defined inside Traits. I was reading some [notes here](http://stackoverflow.com/questions/24357985/php-traits-defining-generic-constants) on this and it sounds like we either need to convert these to static properties (so they can be used inside a Trait) or wrap them in a class. What do you suggest?

raamdev:
@jaswsinc I've finished everything in Step 1 and 1.3, minus the question I had about the two constants files. :smile:

jaswsinc:
> on this and it sounds like we either need to convert these to static properties (so they can be used inside a Trait) or wrap them in a class. What do you suggest?

- Add a new sub-folder [here](https://github.com/websharks/comet-cache-pro/tree/feature/635-traits/src/includes) and name it `interfaces`.

- After [this line](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/composer.json#L46) add the following:

  ```json
  "WebSharks\\CometCache\\Pro\\Interfaces\\": "src/includes/interfaces/",
  ```

- Create a new Interface: `/src/includes/interfaces/Shared/NcDebugConsts.php`

  ```php
  <?php
  namespace WebSharks\CometCache\Pro\Interfaces\Shared;

  interface NcDebugConsts {
      const MY_CONSTANT = 'my value';
  }
  ```

  _Note, you can ditch these lines in the `NcDebugConsts.php` file._
  https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/traits/Shared/CachePathConsts.php#L6-L8
 

- In `Classes\Plugin` and in `Classes\AdvancedCache`, add the `implements` keyword.

  ```php
  use WebSharks\CometCache\Pro\Interfaces; // Comes after the namespace in the file.
  ```

  ```php
  class Plugin implements Interfaces\Shared\NcDebugConsts {
  ```

- Now, search the entire codebase for all instances of those Constants, and prefix them with `$this::`; i.e., `$this:: CACHE_PATH_DEFAULT`. Why? Because they are no longer Namespace Constants, they are now Class Constants, and must be referenced against the current instance that implement them; i.e., `$this::CLASS_CONSTANT`.

---

Another thing you can do is search the Trait and Interface files for instances of docBlock comments that were previously declared with `/*` and change those to the official docBlock syntax now that we have formal methods, properties, constants; i.e., `/**` instead of `/*`.

jaswsinc:
## Step 1.6 of 2 (ha)

- In `Classes\Plugin` and in `Classes\AdvancedCache`, add the `use` keyword.

  ```php
  use WebSharks\CometCache\Pro\Traits; // Comes after the namespace in the file.
  ```

  ```php
  class Plugin implements Interfaces\Shared\NcDebugConsts {

  use Traits\Shared\[Trait Name]; // One for each Shared Trait.
  
  use Traits\Plugin\[Trait Name]; // One for each Plugin Trait.
  ```

- Repeat for the `AdvancedCache` class, but exclude the `\Plugin` Traits and _include_ the `\Ac` Traits.

jaswsinc:
## Step 1.9 of 2

- Remove [these lines](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AdvancedCache.php#L40-L48) from the constructor in `Classes\Plugin` and in `Classes\AdvancedCache`

- Review the entire codebase and look for things like `new ClassName`, `instanceof ClassName`, `ClassName::`, etc. In many cases, you will need to add `use WebSharks\CometCache\Pro\Classes` to the top of the file and prefix the class name with `Classes\`. e.g., https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/plugin.php

 ## Step 2 of 2

- Run a few tests and see what I didn't think of yet. High odds that you will find a few issues; i.e., something that we missed along the way here.

jaswsinc:
- Remove [these lines](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AbsBaseAp.php#L20-L28) also please.

jaswsinc:
- If you want to, you can `use` the Shared Traits just once in this base class that both `Plugin` and `AdvancedCache` extend. https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AbsBaseAp.php#L9

  You will still need to `use` the Ac and Plugin Traits for each class, but you can shave a few bytes off the codebase if you'd like by listing the Shared Traits only once in that base class: `AbsBaseApp.php`.

raamdev:
> Step 1.9 

:laughing: I got excited seeing that, knowing the end _might_ be near.

raamdev:
@jaswsinc writes...
```php
class Plugin implements Interfaces\Shared\NcDebugConsts {

use Traits\Shared\[Trait Name]; // One for each Shared Trait.
```

I get "Undefined class BlogUtils" from the PhpStorm inspector when I add `use Traits\Shared\BlogUtils;`. Should that be normal? (Yes, I already have `use WebSharks\CometCache\Pro\Traits;` at the top of the file underneath the Namespace.)

raamdev:
@jaswsinc Ignore the above. False alarm. The inspector had not updated itself.

jaswsinc:
Cool. Also, I updated the sample I posted above earlier. You probably want something like this.

```php
class Plugin extends AbsBaseApp implements Interfaces\Shared\NcDebugConsts, Interfaces\Shared\CachePathConsts {
```

Or, if you use `AbsBaseApp` to implement the Shared Constants/Traits, you might have:

```php
class Plugin extends AbsBaseApp {
```

raamdev:
@jaswsinc writes...
> - Remove [these lines](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AdvancedCache.php#L40-L48) from the constructor in `Classes\Plugin` and in `Classes\AdvancedCache`
> 
> - Remove [these lines](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AbsBaseAp.php#L20-L28) also please.

Those links are broken after my last few commits, sorry. Can you update them?

jaswsinc:
https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AdvancedCache.php#L72-L80

https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AbsBaseAp.php#L20-L28

jaswsinc:
https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/Plugin.php#L216-L224

jaswsinc:
Another thing you could do (slightly outside the scope of this PR), but as a micro-optimization, would be to look for instances of `dirname(__FILE__)` and change those to `__DIR__`; avoiding the call to `dirname()`.

e.g., `dirname(__FILE__)` becomes `__DIR__`
e.g., `dirname(dirname(__FILE__))` becomes `dirname(__DIR__)`

raamdev:
@jaswsinc Was this right? https://github.com/websharks/comet-cache-pro/commit/83979dd431c0521f9e9292b14477735d615f6bbc I'm running some tests and `$this::CACHE_PATH_REGEX_DEFAULT_SUFFIX_FRAG` as an argument in `src/includes/traits/Shared/CachePathUtils.php` is producing a parse error.

raamdev:
Also: 
> PHP Warning:  Class 'WebSharks\CometCache\Pro\AdvCacheBackCompat' not found in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/stub.php on line 33

raamdev:
Also, lots of PHP Notices about `GLOBAL_NS` and `IS_PRO` like this:

> PHP Notice:  Use of undefined constant GLOBAL_NS - assumed 'GLOBAL_NS' in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/classes/Conflicts.php on line 31

jaswsinc:
> PHP Warning: Class 'WebSharks\CometCache\Pro\AdvCacheBackCompat' not found in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/stub.php on line 33

Add `\Classes` to these: https://github.com/websharks/comet-cache-pro/blob/160227/src/includes/stub.php#L33-L37

jaswsinc:
Nothing that we will need to do a careful review of this file and add class aliases that allow for back compat with existing copies of `advanced-cache.php` during an upgrade to this new version that uses `\Classes` and `\Traits`. https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/templates/advanced-cache.txt

raamdev:
> Add \Classes to these: https://github.com/websharks/comet-cache-pro/blob/160227/src/includes/stub.php#L33-L37

As in this?

```
class_alias(__NAMESPACE__.'\\Classes\\AdvCacheBackCompat', 'WebSharks\\Comet_Cache\\AdvCacheBackCompat');
class_alias(__NAMESPACE__.'\\Classes\\AdvancedCache', 'WebSharks\\Comet_Cache\\AdvancedCache');
```

jaswsinc:
> As in this?

That's it :-)

jaswsinc:
> Also, lots of PHP Notices about GLOBAL_NS and IS_PRO like this:

I see. I didn't think of that. Hmm, this could be a fun one. We need PHP 5.6 to make it easy to fix; i.e., so that we can use those Constants in any file without needing to go through and update the rest of the codebase. I'll have to think about how we could do this with a PHP 5.4 target.

At the moment, it seems that we will need to do something like this, which is not ideal, but perhaps the best we can do in PHP 5.4.

```php
use WebSharks\CometCache\Pro as p; // p stands for plugin.

p\GLOBAL_NS;
```

jaswsinc:
> I'll have to think about how we could do this with a PHP 5.4 target.

Here's an easier way. We can define these constants in each sub-namespace we are using since they are already considered global in the codebase. https://github.com/websharks/comet-cache-pro/blob/160227/src/includes/stub.php#L21-L28

```php
define(__NAMESPACE__.'\\SHORT_NAME', 'CC');
define(__NAMESPACE__.'\\NAME', 'Comet Cache');
define(__NAMESPACE__.'\\DOMAIN', 'cometcache.com');
define(__NAMESPACE__.'\\GLOBAL_NS', 'comet_cache');
define(__NAMESPACE__.'\\SLUG_TD', 'comet-cache');
define(__NAMESPACE__.'\\VERSION', ${__FILE__}['version']);
define(__NAMESPACE__.'\\PLUGIN_FILE', ${__FILE__}['plugin']);
define(__NAMESPACE__.'\\IS_PRO', **${__FILE__}['is_pro']);

foreach (['Classes', 'Traits\\Shared', 'Traits\\Ac', 'Traits\\Plugin', 'Interfaces\\Shared' as ${__FILE__}['_sub_namespace']) {
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\SHORT_NAME', SHORT_NAME);
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\NAME', NAME);
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\DOMAIN', DOMAIN);
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\GLOBAL_NS', GLOBAL_NS);
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\SLUG_TD', SLUG_TD);
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\VERSION', VERSION);
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\PLUGIN_FILE', PLUGIN_FILE);
    define(__NAMESPACE__.'\\'.${__FILE__}['_sub_namespace'].'\\IS_PRO', IS_PRO);
}
```

That should resolve the problem without needing to search/replace other files.

jaswsinc:
- You should also search the entire codebase for any instances of `__NAMESPACE__` and see if they need adjustment. I don't think you will find too many cases of this, because we started using `GLOBAL_NS` in favor of `__NAMESPACE__` a while back. However, it's important to double-check the codebase to see if what _was_ expecting `__NAMESPACE__` to be `WebSharks\CometCache\Pro` is now actually a sub-namespace like `WebSharks\CometCache\Pro\Classes`.

The `__NAMESPACE__` constant will always be equal to the namespace of the file where it is used. Since we altered the Namespace in several files, a review of any uses of `__NAMESPACE__` would be awesome. For instance, you just found a case where it needed to be appended with `\\Classes` in the stub. There may be other cases like that, or of a different variation.

raamdev:
@jaswsinc writes...
> You should also search the entire codebase for any instances of __NAMESPACE__ and see if they need adjustment.

The two places I found were here:

https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/traits/Plugin/CronUtils.php#L37-L60

and here:

https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/api.php

`CronsUtils` looks like `__NAMESPACE__` should be replaced with `GLOBAL_NS`.

`api.php` looks like it should have `use WebSharks\CometCache\Pro\Classes;` added to the top and `class_alias(__NAMESPACE__.'\\ApiBase'` changed to `class_alias(__NAMESPACE__.'\\Classes\ApiBase'`, does that sound right?

jaswsinc:
> https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/traits/Plugin/CronUtils.php#L37-L60
> CronsUtils looks like __NAMESPACE__ should be replaced with GLOBAL_NS.

I think that can stay like it is, but if you do alter it in some way, be sure it identifies the Pro version vs. Lite in some way. The reason we have `__NAMESPACE__` there is because the `__NAMESPACE__` will be different in lite than it is in pro. In this way the CRONs are setup again whenever you install the pro version, which is what we need to be sure of.

jaswsinc:
> and here:
> 
> https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/api.php

It looks like those should be prefixed with `\\Classes`, because ApiBase now lives in the `\Classes` sub-namespace.

jaswsinc:
> class_alias(__NAMESPACE__.'\\Classes\ApiBase', does that sound right?

Sorry. Yes, I agree.

raamdev:
As of now, here's the only error that shows up when I try to install this branch:

```
[28-Feb-2016 21:30:05 UTC] PHP Fatal error:  Uncaught exception 'Exception' with message 'Undefined overload property: `wdRegexToActualRegexFrag`.' in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/classes/AbsBase.php:116
Stack trace:
#0 /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/traits/Shared/PatternUtils.php(22): WebSharks\CometCache\Pro\Classes\AbsBase->__get('wdRegexToActual...')
#1 /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/traits/Plugin/InstallUtils.php(322): WebSharks\CometCache\Pro\Classes\AbsBaseAp->lineDelimitedPatternsToRegex('w3c_validator')
#2 /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/traits/Plugin/InstallUtils.php(258): WebSharks\CometCache\Pro\Classes\Plugin->addAdvancedCache()
#3 [internal function]: WebSharks\CometCache\Pro\Classes\Plugin->checkAdvancedCache('')
#4 /Users/raam/Sites/zcpro.dev/wordpress/wp-includes/plugin.php(525): call_user_func_array(Array, Array)
#5 /Users/raam/Si in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/classes/AbsBase.php on line 116
```

jaswsinc:
I see. Getting close :-) In [this line](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/traits/Shared/PatternUtils.php#L22) we should have:

`[$this, 'wdRegexToActualRegexFrag']` instead of `$this->wdRegexToActualRegexFrag`, because it is a method now instead of a property/Closure.

It might be a good idea to search the rest of the codebase for instances of `array_map($this` and `preg_replace_callback` to see if there are any other cases like this.

raamdev:
Should we not be using anonymous functions anymore, or those are fine? For example here: https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/CdnFilters.php#L364

raamdev:
Reviewed all uses of `array_map()` and `preg_replace_callback()`; nothing else uses `$this`.

jaswsinc:
Cool. Closures are OK to use, yes. However, one way we can improve them slightly is that we can ditch `use($_this)` and also get rid of the `$_this` instance references. Starting w/ PHP 5.4, we can use `$this` inside Closures without needing to `use()` them.

In other words, you can delete [this line](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/CdnFilters.php#L343) and then alter the Closure by removing `use($_this)` altogether. Then, inside the Closure change `$_this` to the standard `$this`.

raamdev:
As of that last commit, Comet Cache Pro builds, installs, and activates without errors, notices, or warnings. Woohoo! :-)

raamdev:
Going to dinner, be back in an hour or so to continue...

jaswsinc:
Yay! Nice teamwork :-)

raamdev:
Loading fine on the backend, but visiting the front-end to see a page get cached results in fatal error:

```
[28-Feb-2016 23:22:42 UTC] PHP Fatal error:  Undefined class constant 'self::CACHE_PATH_DEFAULT' in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/traits/Shared/CachePathUtils.php on line 56
[28-Feb-2016 23:22:42 UTC] PHP Stack trace:
[28-Feb-2016 23:22:42 UTC] PHP   1. {main}() /Users/raam/Sites/zcpro.dev/index.php:0
[28-Feb-2016 23:22:42 UTC] PHP   2. require() /Users/raam/Sites/zcpro.dev/index.php:17
[28-Feb-2016 23:22:42 UTC] PHP   3. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-blog-header.php:12
[28-Feb-2016 23:22:42 UTC] PHP   4. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-load.php:37
[28-Feb-2016 23:22:42 UTC] PHP   5. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-config.php:88
[28-Feb-2016 23:22:42 UTC] PHP   6. include() /Users/raam/Sites/zcpro.dev/wordpress/wp-settings.php:74
[28-Feb-2016 23:22:42 UTC] PHP   7. WebSharks\CometCache\Pro\Classes\AdvancedCache->__construct() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/advanced-cache.php:433
[28-Feb-2016 23:22:42 UTC] PHP   8. WebSharks\CometCache\Pro\Classes\AdvancedCache->maybeStartOutputBuffering() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/classes/AdvancedCache.php:67
[28-Feb-2016 23:22:42 UTC] PHP   9. WebSharks\CometCache\Pro\Classes\AbsBaseAp->buildCachePath() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/traits/Ac/ObUtils.php:189
```

raamdev:
Do I need to `use WebSharks\CometCache\Pro\Interfaces` in `traits/Shared/CachePathUtils.php` to make those constants available?

raamdev:
@jaswsinc Hmm, nope that doesn't work.

raamdev:
Ah, `AbsBase` needs to `use WebSharks\CometCache\Pro\Interfaces;` and then `implements Interfaces\Shared\NcDebugConsts, Interfaces\Shared\CachePathConsts`, right?

raamdev:
In which case, `Plugin` and `AdvancedCache`, which both `extends AbsBaseAp` which itself `extends AbsBase`, does not need to also `implements Interfaces\Shared\NcDebugConsts, Interfaces\Shared\CachePathConsts`. Do I have that right?

raamdev:
@jaswsinc I tried my idea above ↑ and got a different error:

```
[28-Feb-2016 23:39:28 UTC] PHP Fatal error:  Call to undefined function WebSharks\CometCache\Pro\Traits\Ac\__() in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/traits/Ac/ObUtils.php on line 221
[28-Feb-2016 23:39:28 UTC] PHP Stack trace:
[28-Feb-2016 23:39:28 UTC] PHP   1. {main}() /Users/raam/Sites/zcpro.dev/index.php:0
[28-Feb-2016 23:39:28 UTC] PHP   2. require() /Users/raam/Sites/zcpro.dev/index.php:17
[28-Feb-2016 23:39:28 UTC] PHP   3. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-blog-header.php:12
[28-Feb-2016 23:39:28 UTC] PHP   4. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-load.php:37
[28-Feb-2016 23:39:28 UTC] PHP   5. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-config.php:88
[28-Feb-2016 23:39:28 UTC] PHP   6. include() /Users/raam/Sites/zcpro.dev/wordpress/wp-settings.php:74
[28-Feb-2016 23:39:28 UTC] PHP   7. WebSharks\CometCache\Pro\Classes\AdvancedCache->__construct() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/advanced-cache.php:433
[28-Feb-2016 23:39:28 UTC] PHP   8. WebSharks\CometCache\Pro\Classes\AdvancedCache->maybeStartOutputBuffering() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/classes/AdvancedCache.php:65
```

I'm not sure if that means my idea above was wrong or if this is just the next error we need to tackle, but I'm stopping here until I hear from you. :smile: 

---

Also, I'm looking over `traits/Ac/ObUtils.php` and not really seeing how those constants (that ultimately get defined in `wp-content/advanced-cache.php`) become accessible there.

Should all references in the codebase to `COMET_CACHE_*` constants also be updated to use `$this::COMET_CACHE_*`? e.g., `$this::COMET_CACHE_FEEDS_ENABLE`, `$this::COMET_CACHE_GET_REQUESTS`, etc. (outside of `advanced-cache.txt` template of course).

jaswsinc:
> Do I need to use WebSharks\CometCache\Pro\Interfaces in traits/Shared/CachePathUtils.php to make those constants available?

No, you shouldn't need to do that.

> Ah, AbsBase needs to use WebSharks\CometCache\Pro\Interfaces; and then implements Interfaces\Shared\NcDebugConsts, Interfaces\Shared\CachePathConsts, right?

Almost, yes. The problem you're having with the Constants right now is related to [this class](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AbsBaseAp.php) where you are using several Traits that depend on interface constants. While you _are_ using those interface constants, you're not yet using them in the `AbsBaseAp{}` class where the Traits are brought in. So it's confusing PHP a little bit there.

- Change [this line](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AbsBaseAp.php#L11) to:

  ```php
  abstract class AbsBaseAp extends AbsBase implements Interfaces\Shared\NcDebugConsts, Interfaces\Shared\CachePathConsts
  ```

- Add `use WebSharks\CometCache\Pro\Interfaces;` to the [top of that file](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AbsBaseAp.php#L4) also.

- Remove the `implements` portion from [this line](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/Plugin.php#L15).

- Remove it [here](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/classes/AdvancedCache.php#L13) also.

jaswsinc:
- Missing a `p` here I think: https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/traits/Ac/ObUtils.php#L15

jaswsinc:
> Should all references in the codebase to COMET_CACHE_* constants also be updated to use $this::COMET_CACHE_*? e.g., $this::COMET_CACHE_FEEDS_ENABLE, $this::COMET_CACHE_GET_REQUESTS, etc. (outside of advanced-cache.txt template of course).

No, because all of the Constants that start with `COMET_CACHE_` are defined in the global namespace, so they are accessible as-is in all namespaces, including sub-namespaces and in Traits. https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/templates/advanced-cache.txt#L49

jaswsinc:
> In which case, Plugin and AdvancedCache, which both extends AbsBaseAp which itself extends AbsBase, does not need to also implements Interfaces\Shared\NcDebugConsts, Interfaces\Shared\CachePathConsts. Do I have that right?

Yes, that's right. See details above also.

jaswsinc:
> Call to undefined function WebSharks\CometCache\Pro\Traits\Ac\__()

Try changing the namespace in [this file](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/functions/i18n-utils.php) by appending `\Classes`. That's a polyfill for the early phase where the global `__()` is not available yet. It is needed by the `AdvancedCache` class, so it is needs to be defined in the `\Classes` namespace where that class runs.

Loaded here: https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/stub.php#L13

jaswsinc:
_It might be worth your time to write a `.build.php` script in each of the Trait directories, which would be tasked with generating a complete list of the `use Trait\X\X;` lines for you; i.e., do it by scanning the directory of files so it is hard to forget one of these in the future should we decide to add new ones._

_Hopefully in PHP 7.x.x we will get the ability to `use Traits\Shared\*;` or something like that._

raamdev:
@jaswsinc writes...
> Try changing the namespace in [this file](https://github.com/websharks/comet-cache-pro/blob/feature/635-traits/src/includes/functions/i18n-utils.php) by appending `\Classes`.

Hmm, nope, still getting the same error:

```
[29-Feb-2016 11:28:02 UTC] PHP Fatal error:  Call to undefined function WebSharks\CometCache\Pro\Traits\Ac\__() in /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/traits/Ac/ObUtils.php on line 221
[29-Feb-2016 11:28:02 UTC] PHP Stack trace:
[29-Feb-2016 11:28:02 UTC] PHP   1. {main}() /Users/raam/Sites/zcpro.dev/index.php:0
[29-Feb-2016 11:28:02 UTC] PHP   2. require() /Users/raam/Sites/zcpro.dev/index.php:17
[29-Feb-2016 11:28:02 UTC] PHP   3. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-blog-header.php:12
[29-Feb-2016 11:28:02 UTC] PHP   4. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-load.php:37
[29-Feb-2016 11:28:02 UTC] PHP   5. require_once() /Users/raam/Sites/zcpro.dev/wordpress/wp-config.php:88
[29-Feb-2016 11:28:02 UTC] PHP   6. include() /Users/raam/Sites/zcpro.dev/wordpress/wp-settings.php:74
[29-Feb-2016 11:28:02 UTC] PHP   7. WebSharks\CometCache\Pro\Classes\AdvancedCache->__construct() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/advanced-cache.php:433
[29-Feb-2016 11:28:02 UTC] PHP   8. WebSharks\CometCache\Pro\Classes\AdvancedCache->maybeStartOutputBuffering() /Users/raam/Sites/zcpro.dev/wordpress/wp-depo/plugins/comet-cache-pro/src/includes/classes/AdvancedCache.php:65
```

jaswsinc:
Hmm. I see. What is missing is `__()` in the Trait namespace where it called upon.

Try this in the i18n utils file; i.e., make the polyfill available in sub-namespaces also.

```php
<?php
namespace WebSharks\CometCache\Pro
{
    /**
     * Polyfill for {@link \__()}.
     *
     * @since 150422 Rewrite.
     *
     * @param string $string      String to translate.
     * @param string $text_domain Plugin text domain.
     *
     * @return string Possibly translated string.
     */
    function __($string, $text_domain)
    {
        static $exists; // Cache.

        if ($exists || ($exists = function_exists('__'))) {
            return \__($string, $text_domain);
        }
        return $string; // Not possible (yet).
    }
}
namespace WebSharks\CometCache\Pro\Traits\Ac
{
    function __($string, $text_domain)
    {
        return \WebSharks\CometCache\Pro\__($string, $text_domain);
    }
}
namespace WebSharks\CometCache\Pro\Traits\Shared
{
    function __($string, $text_domain)
    {
        return \WebSharks\CometCache\Pro\__($string, $text_domain);
    }
}
```

raamdev:
@jaswsinc WOohooO! As of now, no errors, notices, or warnings while installing, activating, enabling, changing/saving options, or caching on the front-end. Also tested with the HTML Compressor enabled and that also works great. 

If that's all we need to do for the conversion to Traits, I'd say this is ready for some quick testing by you and then a merge into `000000-dev` to continue with https://github.com/websharks/comet-cache/issues/635.

raamdev:
> It might be worth your time to write a .build.php script in each of the Trait directories, which would be tasked with generating a complete list of the use Trait\X\X; lines for you; i.e., do it by scanning the directory of files so it is hard to forget one of these in the future should we decide to add new ones.

Can this be automated with Phing?

jaswsinc:
> If that's all we need to do for the conversion to Traits, I'd say this is ready for some quick testing by you and then a merge into 000000-dev to continue with websharks/comet-cache#635.

Great! Nice work.

jaswsinc:
> Can this be automated with Phing?

Yes, but I'd still suggest writing it in PHP and just name the file `.build.php`. In the Phing build script we can write a directory scanner that will automatically execute those sub-routines in any file with the name `.build.php` later. For now, you can just run `.build.php` to generate the list of Traits manually.

raamdev:
@jaswsinc writes...
> I'd still suggest writing it in PHP and just name the file .build.php.

Done. See https://github.com/websharks/comet-cache-pro/commit/be95571b334393d5b066d610e1e48ce02266e4a6.

Do you think this is ready-to-merge yet?

PR #236, opened 2016-04-08
Pro Preview Manual Cache Clearing panel has the wrong color; 

renzms:
see: websharks/comet-cache#711


src/includes/classes/MenuPageOptions.php
```diff
@@ -251,9 +251,9 @@ public function __construct()
         /* --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- */
 
         if (IS_PRO || $this->plugin->isProPreview()) {
-            echo '<div class="plugin-menu-page-panel'.(!IS_PRO ? ' pro-preview' : '').'">'."\n";
```
jaswsinc: @renzms I think this line should be restored to what it was. It still applies, even though the additional class is added to the header for this panel, we still flag the entire panel as being a part of the `pro-preview`.

raamdev: @jaswsinc Actually, the way this worked changed awhile back, as part of the improvements to the Pro Preview. What @renzms did here is correct.

PR #237, opened 2016-04-08
PR: feature/464

jaswsinc:
See: websharks/comet-cache#464


jaswsinc:
Referencing: https://github.com/wp-cli/wp-cli/blob/master/php/wp-cli.php#L4

jaswsinc:
@raamdev This is ready to review.

raamdev:
@jaswsinc Could you please explain [the changes](https://github.com/websharks/comet-cache-pro/pull/237/commits/5abc1008e20311fb994a15bb5cb59698b8aa9062) made to `tests/includes/phpunit.php` in this PR for me?

jaswsinc:
Sure. On Travis CI, the plugin is automatically activated via WP CLI. In the past, activating the plugin via WP CLI didn't quite do the trick, because you also had to enable Comet Cache from the Dashboard. So those additional lines in our PHPUnit bootstrap were enabling the plugin so that it would be fully active for our tests. With the changes in this PR, activating the plugin with WP CLI now fully enables the plugin, which means those lines in our PHPUnit bootstrap are no longer necessary.

raamdev:
Ah, that explains it! Thank you. :-)

PR #239, opened 2016-04-09
Try Pro Updater proxy server when primary fails

raamdev:
See websharks/comet-cache#678


raamdev:
This PR also updates the inline docs to explain that a license key is required (removes all mention of using an account password for authentication). See https://github.com/websharks/cometcache.com/issues/64

raamdev:
@jaswsinc A second set of eyes on this would be appreciated, when you get a chance. Thank you.

jaswsinc:
Looks good to me 😃

PR #242, opened 2016-04-12
Attempt to use proxy server when checks for latest version fail primary

raamdev:
See websharks/comet-cache#678


raamdev:
@jaswsinc Just a heads up on these changes. 

During testing I found that the proxy server was not always responding quickly enough and the default `wp_remote_post()` timeout was causing failures ("Operation timed out after 1001 milliseconds with 0 bytes received").

jaswsinc:
Copy that. I thought the default was `5` seconds, but I have seen that sort of thing too. I still think there is a bug in WordPress core that causes this at times. I've never been able to put my finger on it though.

PR #243, opened 2016-04-14
Enable Comet Cache to clear cache for Custom Term Archive Views by default

renzms:
see websharks/comet-cache#693


jaswsinc:
This looks great to me. 😄

raamdev:
@jaswsinc This will change the default for new users, but not for existing users, correct? For that we also need a Version Upgrade routine that changes the value of this option from `0` to `1`. Do I have that right?

jaswsinc:
Well, I'm not sure how that would work exactly though, because it's just a boolean flag. I can't think of a way to do a version-specific upgrade for this without overriding (potentially) someone's personal preference of `0` as it is currently configured by their options. If we want to just enable it for everyone during the next update and then post a notice though, I suppose that would work.

raamdev:
> I can't think of a way to do a version-specific upgrade for this without overriding (potentially) someone's personal preference of 0 as it is currently configured by their options.

Well, it's off by default right now, so someone's personal preference would really be "leave it at the default". If we decide to change the default, then we're still going with their personal preference: use the default.

In my mind, it's a lot safer to enable this for everybody because the most likely scenario is that site owners don't even realize the Custom Term Archive Views are not getting cleared automatically. The much less likely scenario is that someone is specifically relying on that option being disabled for some other part of their site to work correctly. 

And even then, "not working correctly" would most likely mean the term archives view caches getting cleared unnecessarily (e.g., if a plugin that adds a CPT with a Custom Term calls the wrong hooks at the wrong times and results in Comet Cache clearing the CPT Term Archive View cache too frequently).

jaswsinc:
Agreed 👍

raamdev:
@jaswsinc A quick review of https://github.com/websharks/comet-cache-pro/pull/243/commits/f22e7ba9d3fb7c47a38d9e416492a82f599487ca when you get a chance, please.

jaswsinc:
Looks great 😄

renzms:
@raamdev @jaswsinc 

Hi, noticed this was still open, anything else I need to add here or fix? Thanks!

raamdev:
@renzms Thanks for the heads up. :-) This has been merged.

PR #246, opened 2016-04-28
Clear the PHP OPCache when a plugin is upgraded/activated/deactivated

raamdev:
See websharks/comet-cache#740

### DIFF
--- a/src/includes/classes/Actions.php
+++ b/src/includes/classes/Actions.php
@@ -805,8 +805,8 @@ class Actions extends AbsBase
                          '_wpnonce' => wp_create_nonce('upgrade-plugin_'.plugin_basename(PLUGIN_FILE)),
 
                          // See: `preSiteTransientUpdatePlugins()` where these are picked up.
-                         GLOBAL_NS.'_update_pro_version' => $product_api_response->pro_version,
-                         GLOBAL_NS.'_update_pro_zip'     => base64_encode($product_api_response->pro_zip),
+                         GLOBAL_NS.'_update_pro_version' => apply_filters(GLOBAL_NS.'_update_pro_version', $product_api_response->pro_version),
+                         GLOBAL_NS.'_update_pro_zip'     => base64_encode(apply_filters(GLOBAL_NS.'_update_pro_zip', $product_api_response->pro_zip)),
                          // @TODO Encrypt/decrypt to avoid mod_security issues. Base64 is not enough.
         ];
         $redirect_to = add_query_arg(urlencode_deep($query_args), $redirect_to);
--- a/src/includes/classes/Plugin.php
+++ b/src/includes/classes/Plugin.php
@@ -525,6 +525,7 @@ class Plugin extends AbsBaseAp
         add_action('admin_init', [$this, 'autoClearCacheOnSettingChanges']);
         add_action('safecss_save_pre', [$this, 'autoClearCacheOnJetpackCustomCss'], 10, 1);
         add_action('upgrader_process_complete', [$this, 'autoClearOnUpgraderProcessComplete'], 10, 2);
+        add_action('upgrader_process_complete', [$this, 'wipeOpcacheByForce'], PHP_INT_MAX);
 
         add_action('switch_theme', [$this, 'autoClearCache']);
         add_action('wp_create_nav_menu', [$this, 'autoClearCache']);
--- a/src/includes/traits/Plugin/WcpOpcacheUtils.php
+++ b/src/includes/traits/Plugin/WcpOpcacheUtils.php
@@ -64,6 +64,18 @@ trait WcpOpcacheUtils
     }
 
     /**
+     * Wipe the Opcache (by force).
+     *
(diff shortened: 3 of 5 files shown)


raamdev:
@jaswsinc As I mentioned on Slack, what I've pushed so far doesn't seem to clear the PHP OPCache after an update via the Pro Updater. I'm not sure why—still trying to figure that out. I tested regular WordPress plugin updates and the OPCache is wiped right after as expected, but running the Pro Updater does not seem to trigger `wipeOpcacheByForce()` (which is being attached to the WordPress `shutdown` hook). 

Maybe you can see something I'm missing? 😄

raamdev:
**Note:** While https://github.com/websharks/comet-cache-pro/pull/246/commits/a75f879645e4cb64476d0904da77dc1f239ea349 causes the Pro Updater to clear the PHP OPCache after installing an updated version of Comet Cache Pro, there's still a problem: the `add_action()` on [these lines](https://github.com/websharks/comet-cache-pro/pull/246/commits/9f096e380a401cb8528b9463ae2f6c064252f889) should be causing a Pro Update to attach `wipeOpcacheByForce()` to the WordPress `shutdown` hook, but that's not happening.

Here's why:

I did some XDebugging and found that [this line](https://github.com/websharks/comet-cache-pro/blob/160417/src/includes/traits/Plugin/WcpUpdaterUtils.php#L49) `is_plugin_active($data['plugin'])` is returning false (with `$data['plugin'] == 'comet-cache-pro/comet-cache-pro.php`). That's preventing the following line from being run whenever the Pro Updater is being used to update Comet Cache Pro and therefore this line never gets run:

```
add_action('shutdown', [$this, 'wipeOpcacheByForce'], PHP_INT_MAX);
```

raamdev:
Here's what's going on: 

When upgrading an active Comet Cache Pro, the `upgrade()` routine in `class-wp-upgrader.php` is [deactivating the plugin](https://github.com/WordPress/WordPress/blob/4.5.1/wp-admin/includes/class-wp-upgrader.php#L976) before running the update that eventually calls our `autoClearOnUpgraderProcessComplete()` routine, so by the time the routine runs, `is_plugin_active($data['plugin'])` returns false.

When upgrading other active WordPress plugins from the WordPress Dashboard, they are always updated as a "bulk" upgrade, even if you're only updating one plugin. A different routine (`bulk_upgrade()`) is used in this case and in that routine the plugin [is not deactivated](https://github.com/WordPress/WordPress/blob/4.5.1/wp-admin/includes/class-wp-upgrader.php#L1033), so when our `autoClearOnUpgraderProcessComplete()` routine runs `is_plugin_active($data['plugin'])` returns true, as we're expecting.

raamdev:
@jaswsinc I spent some time going through this and looking for the best way to handle the active plugin conditionals in `autoClearOnUpgraderProcessComplete()`. 

The Pro Updater uses the WordPress `Plugin_Upgrader->upgrade()` routine. All other WordPress plugins and themes use the `Plugin_Upgrader->bulk_upgrade()` routine, even when you're only updating a single plugin/theme. I looked into updating the Pro Updater to use the `bulk_upgrade()` routine instead, but that doesn't look feasible—there would be too much 'working around' behavior intended to for the WordPress Dashboard.

So I eventually ended up with https://github.com/websharks/comet-cache-pro/pull/246/commits/3b4fe7aff687a065376f88b383a58b36849dcf03, which I tested with regular WordPress plugin updates (which always go through `autoClearOnUpgraderProcessComplete()` as `$multi_plugin_update = true`) and with the Pro Updater (which goes through as `$single_plugin_update = true`). It works great. 

_Note: Checking `$upgrader_instance->skin->upgrader->skin->plugin_active` is reliable when doing a single plugin update, but cannot be used when doing a multi-plugin update, hence the reason I didn't just refactor `autoClearOnUpgraderProcessComplete()` to rely on that for checking the plugin active status._

During testing, I also tried temporarily removing [this hook](https://github.com/websharks/comet-cache-pro/pull/246/commits/a75f879645e4cb64476d0904da77dc1f239ea349) to confirm that `wipeOpcacheByForce()` gets called by the `shutdown` hook after running a Pro Update (i.e., the original mystery we were trying to solve here).

----

If you don't have any objections or other suggested changes, I consider this ready to merge. :-) If you could run a few tests for me against this branch, that would be great.

Remember, with this branch you can now add an MU-Plugin like the following to force the Pro Updater to download a specific update (you'll still need to authenticate with the update server):

```php
<?php
add_filter( 'comet_cache_update_pro_version', function() {
    return '160428.740';
}, 10);
add_filter( 'comet_cache_update_pro_zip', function() {
    return 'http://local.wordpress.dev/wp-content/uploads/2016/04/comet-cache-pro-740.zip';
}, 10);
```

raamdev:
@jaswsinc Ping. ↑ (Trying to put together and RC today and would like to get this in.)

jaswsinc:
@raamdev I ran this branch through some tests and no issues. Looks great!

~ Thanks for the MU plugin. 💯

PR #248, opened 2016-05-10
Implement interfaces in AbsBase instead of AbsBaseAp

raamdev:
See websharks/comet-cache#752


raamdev:
@jaswsinc A quick review here when you get a chance.

jaswsinc:
Looks great to me! :-)

PR #250, opened 2016-05-15
Add new UI Config panel for Host Exclusion Patterns; See: websharks/comet-cache#754

kristineds:
Add new UI Config panel for Host Exclusion Patterns; 

screen shot 2016-05-16 at 12 58 01 am

See: websharks/comet-cache#754


kristineds:
@jaswsinc @raamdev Need some help with the wording and the conditional for the `is_multisite()`. https://github.com/websharks/comet-cache-pro/pull/250/commits/266cdf59bf41aeca0a2e2bc718a0f392b9e7d7d9#diff-7303c30aabc49958b1963a5f09c57332R700

jaswsinc:
Lookin' great! :-)

I suggest these opening paragraphs in this panel:

> If there are specific domains that should not be cached, you can enter them here so they are excluded automatically. The easiest way to exclude a host is to enter the full domain name on a line of it's own in the field below.

> This field also supports [Watered-Down Regex](https://cometcache.com/kb-article/watered-down-regex-syntax/) syntax, which means that you can also exclude a pattern like: `*.example.com` or `*.example.*`. So for instance, if you wanted to exclude all child sites and only cache pages on the Main Site of a Network installation, you could exclude all sub-domains using: `*.mynetwork.com`. That excludes `anything.`mynetwork.com, but not `mynetwork.com` by itself.

I would ditch the **Tip:** underneath the panel, but leave the **Note:** as-is.

src/includes/classes/MenuPageOptions.php
```diff
@@ -697,6 +697,28 @@ public function __construct()
 
         echo '<div class="plugin-menu-page-panel">'."\n";
 
+        if (is_multisite() && $this->plugin->applyWpFilters(GLOBAL_NS.'_exclude_hosts_option_enable', is_multisite())) {
```
jaswsinc: This line should be:

```php
if ($this->plugin->applyWpFilters(GLOBAL_NS.'_exclude_hosts_option_enable', is_multisite())) {
```

Meaning, the filter is `true` on a network by default, and the panel is shown. On a standard install, `is_multisite()` is false and the filter will default to being off. In either case, a developer can override by attaching to and altering the filtered value used in the conditional.

kristineds:
@jaswsinc @raamdev Updated the description and conditional (https://github.com/websharks/comet-cache-pro/pull/250/commits/0d45adf78c1e9cd468b24122d1f8bccfdc253a97
) per Jason's feedback. However, when I tried testing, it doesn't seem to work. When I added `*.websharks-inc.net` on Host Exclusion Patterns on my multisite install, that setting doesn't get saved. Any ideas why? :)

screen shot 2016-05-21 at 5 24 28 pm

jaswsinc:
@kristineds I suggest adding a logging routine while you're debugging this so that you can see more about what's happening internally. For instance, in the [updateOptions()](https://github.com/websharks/comet-cache-pro/blob/000000-dev/src/includes/traits/Plugin/OptionUtils.php#L53) routine you could log what occurs there and take a closer look.

```php
file_put_contents(WP_CONTENT_DIR.'/debug.log', print_r(get_defined_vars(), true)."\n\n", FILE_APPEND);
```

---

Instead of `anything.mynetwork.com` I would try:

`anything.`mynetwork.com ; i.e., `anything.` in `<code>` tags instead of the full host name.

just to help improve clarity with respect to `anything.` being the important difference.

---

The [`echo '<div class="plugin-menu-page-panel">'."\n";` line](https://github.com/websharks/comet-cache-pro/blob/0d45adf78c1e9cd468b24122d1f8bccfdc253a97/src/includes/classes/MenuPageOptions.php#L698) should go _inside_ of the conditional that checks if that markup should be created or not. If not, that line should not be printed. The same with the `</div>` that closes that tag.

kristineds:
> For instance, in the updateOptions() routine you could log what occurs there and take a closer look.

@jaswsinc @raamdev Added the debugging routine here. https://github.com/websharks/comet-cache-pro/pull/250/commits/bac2103d5f83b80db54a738a92b602e391e31124

Tested this on a multisite, but the domain still doesn't get save on the field when I save the options. Looking at my `debug.log` file, I'm seeing this PHP Notice:

```
[23-May-2016 10:13:09 UTC] PHP Notice:  Undefined index: exclude_hosts in /src/includes/classes/MenuPageOptions.php on line 716
```

Could you point me in the right direction on what I'm doing wrong? :) 

----------------

> Instead of `anything.mynetwork.com` I would try:
`anything.`mynetwork.com ; i.e., `anything.` in `<code>` tags instead of the full host name.

Updated here. https://github.com/websharks/comet-cache-pro/pull/250/commits/8af593753c9b54812cfe857de647501a1570c148

screen shot 2016-05-23 at 6 26 58 pm

src/includes/classes/VsUpgrades.php
```diff
@@ -94,6 +94,9 @@ protected function fromLte151107()
                 if (!empty($existing_options['cache_clear_xml_sitemap_patterns']) && strpos($existing_options['cache_clear_xml_sitemap_patterns'], '**') === false) {
                     $this->plugin->options['cache_clear_xml_sitemap_patterns'] = str_replace('*', '**', $existing_options['cache_clear_xml_sitemap_patterns']);
                 }
+                if (!empty($existing_options['exclude_hosts']) && strpos($existing_options['exclude_hosts'], '**') === false) {
```
jaswsinc: You won't need these lines, because this option did not exist prior; i.e., nothing to upgrade.

kristineds:
> You won't need these lines, because this option did not exist prior; i.e., nothing to upgrade.

@jaswsinc That has been updated. https://github.com/websharks/comet-cache-pro/pull/250/commits/1f3a9aa7cd0cf035dccfe33d2fa8fa5a35f78baf cc @raamdev

raamdev:
@kristineds Awesome work! 😀  I just put this through a couple of tests and it's working GREAT.

@jaswsinc I noticed that the Host Exclusion Patterns panel shows up even when the Multisite Network is configured to use sub-directories (`define('SUBDOMAIN_INSTALL', false);`). Is the Host Exclusions Panel of any use when that's the case, or should it only appear when using sub-domains (`define('SUBDOMAIN_INSTALL', true);`)?

Also, just to confirm: When using multisite with sub-directories, the existing URI Exclusion Patterns feature should be used, correct? (I confirmed that it works with sub-directories; I just want to confirm that's expected.)

If so, I think we should do an additional things there:

- Have a special Multisite Network note appear in URI Exclusion Patterns when `define('SUBDOMAIN_INSTALL', false);` that explains to Network Admins that they can also use URI Exclusion Patterns to exclude specific sites from being cached.

jaswsinc:
> Is the Host Exclusions Panel of any use when that's the case, or should it only appear when using sub-domains (define('SUBDOMAIN_INSTALL', true);)?

Not really, no. An exception might be for sites running a domain-mapping plugin though.

I would show that panel if ...

```php
if ((defined('SUBDOMAIN_INSTALL') && SUBDOMAIN_INSTALL) || $this->plugin->canConsiderDomainMapping()) {

}
```

> Also, just to confirm: When using multisite with sub-directories, the existing URI Exclusion Patterns feature should be used, correct? (I confirmed that it works with sub-directories; I just want to confirm that's expected.)

Correct, yes.

> Have a special Multisite Network note appear in URI Exclusion Patterns when define('SUBDOMAIN_INSTALL', false); that explains to Network Admins that they can also use URI Exclusion Patterns to exclude specific sites from being cached.

Agree.

jaswsinc:
Referencing: https://github.com/websharks/comet-cache-pro/blob/160521/src/includes/traits/Shared/DomainMappingUtils.php#L17

jaswsinc:
@kristineds So, in the context of these changes where we have the filter. The filter could look like this.

```php
$exclude_hosts_option_enable = is_multisite() &&
    ((defined('SUBDOMAIN_INSTALL') && SUBDOMAIN_INSTALL) || $this->plugin->canConsiderDomainMapping());
if ($this->plugin->applyWpFilters(GLOBAL_NS.'_exclude_hosts_option_enable', $exclude_hosts_option_enable)) {
    ...
}
```

kristineds:
@jaswsinc Like this? https://github.com/websharks/comet-cache-pro/pull/250/commits/7516eb94aad4088fdf65765cbf08fa12be0815d4

jaswsinc:
Perfect, yes.

raamdev:
@kristineds Is this ready for another review, or are you still working on this?

kristineds:
@raamdev This is ready for review. :)

src/includes/traits/Plugin/OptionUtils.php
```diff
@@ -62,6 +62,8 @@ public function updateOptions(array $options)
         $this->options = array_intersect_key($this->options, $this->default_options);
         update_site_option(GLOBAL_NS.'_options', $this->options);
 
+        file_put_contents(WP_CONTENT_DIR.'/debug.log', print_r(get_defined_vars(), true)."\n\n", FILE_APPEND);
```
jaswsinc: @kristineds Please remove this line before submitting a PR. This is for _local_ debugging only and should never be included in any merge into the dev branch where it could end up on a live site and create a huge log file over time.

kristineds:
@jaswsinc Yikes. I have removed that line: https://github.com/websharks/comet-cache-pro/pull/250/commits/7bce007f66dfaafaa8b69552e48d00c6c656fcf2. Thanks for checking! cc @raamdev

raamdev:
@kristineds I just ran this through a few more tests. The automatic hiding of the new Host Exclusion Patterns works great.

There's only one thing left to do:

- Have a special Multisite Network note appear at the bottom of URI Exclusion Patterns when `is_multisite() && define('SUBDOMAIN_INSTALL', false);` that explains to Network Admins that they can also use URI Exclusion Patterns to exclude specific sites from being cached. It should appear at the bottom of that panel and start with "**Multisite Networks w/ Sub-Directories:**" and explain that URI Exclusion Patterns can also be used to exclude specific sites, e.g., `/site1/*`.

kristineds:
> Have a special Multisite Network note appear at the bottom of URI Exclusion Patterns when is_multisite() && define('SUBDOMAIN_INSTALL', false); 

@raamdev Like this? https://github.com/websharks/comet-cache-pro/pull/250/commits/93e48bf3f606f1ac4cfdabde4d6020cea51e5f01

raamdev:
@kristineds I tested this (did you test after committing your change?) and I'm getting a notice at the bottom:



It looks like there was a bug in my code example. I suggest searching elsewhere in the codebase for a similar check to see how it should be done and then running a test to make sure the code works as expected.

kristineds:
@jaswsinc @raamdev It seems to work but only when `define('SUBDOMAIN_INSTALL')`. https://github.com/websharks/comet-cache-pro/pull/250/commits/2a3bf734a707cd862be8184677d834e4c9ca2c5b

screen shot 2016-06-11 at 1 51 36 am

raamdev:
@kristineds [`define()`](http://php.net/manual/en/function.define.php) is the wrong function to use here.

raamdev:
@kristineds I see in https://github.com/websharks/comet-cache-pro/pull/250/commits/2a3bf734a707cd862be8184677d834e4c9ca2c5b that you're not using `define()` as your last comment mentions, but rather `defined()`, which _is_ the correct function.

However, `if(defined('SUBDOMAIN_INSTALL') && SUBDOMAIN_INSTALL)` checks if `SUBDOMAIN_INSTALL` is defined and if `SUBDOMAIN_INSTALL` is true. Do we want `SUBDOMAIN_INSTALL` to be true when showing the note about sub-directories?

raamdev:
@kristineds So the questions that need to be asked here are:

- What are we trying to do?
- What condition should this conditional be checking for?

kristineds:
> Do we want SUBDOMAIN_INSTALL to be true when showing the note about sub-directories?

@raamdev Should it be `if(defined('SUBDOMAIN_INSTALL') && !SUBDOMAIN_INSTALL)` instead since what we want is to have `SUBDOMAIN_INSTALL` to be `false` to show the note about sub-directories?

raamdev:
@kristineds writes... 
> Should it be if(defined('SUBDOMAIN_INSTALL') && !SUBDOMAIN_INSTALL)

Yes, that looks correct to me. If you could test that on a multisite network with sub-directories (see https://codex.wordpress.org/Multisite_Network_Administration#Switching_network_types), that would be great. The note should NOT show up when a Multisite Network is using Sub-Domains, since that note is only applicable when using Sub-Directories.

raamdev:
@kristineds FYI, I just merged the latest from `000000-dev` into this feature branch, so you'll want to `git pull` to make sure you have those changes locally if you're still doing testing for this feature branch.

raamdev:
@kristineds FYI, you had the note about Multisite Networks w/ Sub-Directories in the wrong panel; it should've gone in the URI Exclusion Patterns panel, not the Host Exclusion Patterns (in the scenario in which that note appears, the Host Exclusion Patterns panel is hidden, so it would never be seen). I fixed this in https://github.com/websharks/comet-cache-pro/pull/250/commits/c181f9cc0d95530e5d3852e67b1305abd600f7cd.

kristineds:
@raamdev Yikes. 😥  Thanks for catching that.

PR #251, opened 2016-05-17
Dim additional options for Logged-In Users when feature is disabled

raamdev:
See websharks/comet-cache#768


raamdev:
Note this is currently broken; see https://github.com/websharks/comet-cache/issues/768#issuecomment-219787999

PR #253, opened 2016-05-21
Use WP Transients to store Pro Plugin Updater metadata during update

raamdev:
See websharks/comet-cache#416

### DIFF
--- a/src/includes/classes/Actions.php
+++ b/src/includes/classes/Actions.php
@@ -798,17 +798,20 @@ class Actions extends AbsBase
         ]);
         $this->plugin->dismissMainNotice('new-pro-version-available');
 
-        $redirect_to = self_admin_url('/update.php');
-        $query_args  = [ // Like a normal WP plugin.
-                         'action'   => 'upgrade-plugin',
-                         'plugin'   => plugin_basename(PLUGIN_FILE),
-                         '_wpnonce' => wp_create_nonce('upgrade-plugin_'.plugin_basename(PLUGIN_FILE)),
-
-                         // See: `preSiteTransientUpdatePlugins()` where these are picked up.
-                         GLOBAL_NS.'_update_pro_version' => apply_filters(GLOBAL_NS.'_update_pro_version', $product_api_response->pro_version),
-                         GLOBAL_NS.'_update_pro_zip'     => base64_encode(apply_filters(GLOBAL_NS.'_update_pro_zip', $product_api_response->pro_zip)),
-                         // @TODO Encrypt/decrypt to avoid mod_security issues. Base64 is not enough.
+        $update_pro_version = $this->plugin->applyWpFilters(GLOBAL_NS.'_update_pro_version', $product_api_response->pro_version);
+        $update_pro_zip     = $this->plugin->applyWpFilters(GLOBAL_NS.'_update_pro_zip', $product_api_response->pro_zip);
+        $redirect_to        = self_admin_url('/update.php');
+        $query_args         = [ // Like a normal WP plugin.
+                                'action'                        => 'upgrade-plugin',
+                                'plugin'                        => plugin_basename(PLUGIN_FILE),
+                                '_wpnonce'                      => wp_create_nonce('upgrade-plugin_'.plugin_basename(PLUGIN_FILE)),
+                                GLOBAL_NS.'_update_pro_version' => $update_pro_version,
         ];
+
+        // Store the Pro Zip URL in Transient with a 5 minute expiration
+        // See: `preSiteTransientUpdatePlugins()` where these are picked up.
+        set_site_transient(GLOBAL_NS.'_update_pro_zip_'.$update_pro_version, $update_pro_zip, 60*5);
+
         $redirect_to = add_query_arg(urlencode_deep($query_args), $redirect_to);
 
         wp_redirect($redirect_to).exit();
--- a/src/includes/traits/Plugin/UpdateUtils.php
+++ b/src/includes/traits/Plugin/UpdateUtils.php
@@ -128,12 +128,15 @@ trait UpdateUtils
         if (empty($_r['_wpnonce']) || !wp_verify_nonce((string) $_r['_wpnonce'], 'upgrade-plugin_'.plugin_basename(PLUGIN_FILE))) {
             return $transient; // Nothing to do here.
         }
-        if (empty($_r[GLOBAL_NS.'_update_pro_version']) || empty($_r[GLOBAL_NS.'_update_pro_zip'])) {
+        if (empty($_r[GLOBAL_NS.'_update_pro_version'])) {
             return $transient; // Nothing to do here.
         }
+
         $update_pro_version = (string) $_r[GLOBAL_NS.'_update_pro_version'];
-        $update_pro_zip     = base64_decode((string) $_r[GLOBAL_NS.'_update_pro_zip'], true);
-        // @TODO Encrypt/decrypt to avoid mod_security issues. Base64 is not enough.
+
+        if (!($update_pro_zip = get_site_transient(GLOBAL_NS.'_update_pro_zip_'.$update_pro_version))) {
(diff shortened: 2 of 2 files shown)


raamdev:
@jaswsinc As noted in https://github.com/websharks/comet-cache/issues/416#issuecomment-220803064, this currently does not work as expected. I get "The plugin is at the latest version." Not sure why.

raamdev:
Using XDebug, this is where the failure occurs:



So it appears that for some reason [this code](https://github.com/websharks/comet-cache-pro/blob/6382ff6bf591c883e35ebf930236386cc16e7459/src/includes/traits/Plugin/UpdateUtils.php#L147-L155) isn't creating the object... or something.

src/includes/classes/Actions.php
```diff
@@ -798,17 +798,23 @@ protected function proUpdate($args)
...
-        $query_args  = [ // Like a normal WP plugin.
-                         'action'   => 'upgrade-plugin',
-                         'plugin'   => plugin_basename(PLUGIN_FILE),
-                         '_wpnonce' => wp_create_nonce('upgrade-plugin_'.plugin_basename(PLUGIN_FILE)),
-
-                         // See: `preSiteTransientUpdatePlugins()` where these are picked up.
-                         GLOBAL_NS.'_update_pro_version' => apply_filters(GLOBAL_NS.'_update_pro_version', $product_api_response->pro_version),
-                         GLOBAL_NS.'_update_pro_zip'     => base64_encode(apply_filters(GLOBAL_NS.'_update_pro_zip', $product_api_response->pro_zip)),
-                         // @TODO Encrypt/decrypt to avoid mod_security issues. Base64 is not enough.
+        $update_pro_version = apply_filters(GLOBAL_NS.'_update_pro_version', $product_api_response->pro_version);
```
jaswsinc: I suggest using `$this->plugin->applyWpFilters()` here instead.

src/includes/classes/Actions.php
```diff
@@ -798,17 +798,23 @@ protected function proUpdate($args)
...
+                                'plugin'                        => plugin_basename(PLUGIN_FILE),
+                                '_wpnonce'                      => wp_create_nonce('upgrade-plugin_'.plugin_basename(PLUGIN_FILE)),
+                                GLOBAL_NS.'_update_pro_version' => $update_pro_version,
+                                GLOBAL_NS.'_update_pro_zip'     => base64_encode(apply_filters(GLOBAL_NS.'_update_pro_zip', $product_api_response->pro_zip)),
         ];
+
+        // Store the Pro Zip URL in Transient with a 5 minute expiration
+        // See: `preSiteTransientUpdatePlugins()` where these are picked up.
+        if (false === (get_site_transient(GLOBAL_NS.'_update_pro_zip_'.$update_pro_version))) {
+            set_site_transient(GLOBAL_NS.'_update_pro_zip_'.$update_pro_version, $update_pro_zip, 60*5);
```
jaswsinc: I suggest removing the strict comparison here, as it is difficult to rely on WordPress to adhere to strict types; i.e., even if this works now, it could break in the future. The WordPress codebase is improving in this regard, but there are still a lot of parameters and a lot of functions that can return a wide variety of data types.

src/includes/classes/Actions.php
```diff
@@ -798,17 +798,23 @@ protected function proUpdate($args)
...
+                                'action'                        => 'upgrade-plugin',
+                                'plugin'                        => plugin_basename(PLUGIN_FILE),
+                                '_wpnonce'                      => wp_create_nonce('upgrade-plugin_'.plugin_basename(PLUGIN_FILE)),
+                                GLOBAL_NS.'_update_pro_version' => $update_pro_version,
+                                GLOBAL_NS.'_update_pro_zip'     => base64_encode(apply_filters(GLOBAL_NS.'_update_pro_zip', $product_api_response->pro_zip)),
         ];
+
+        // Store the Pro Zip URL in Transient with a 5 minute expiration
+        // See: `preSiteTransientUpdatePlugins()` where these are picked up.
+        if (false === (get_site_transient(GLOBAL_NS.'_update_pro_zip_'.$update_pro_version))) {
```
jaswsinc: Also, I wouldn't worry about checking if it exists here, because this routine is designed specifically to update those values with the latest details that it just pulled from the API. So I'd skip the `get_site_transient()` call altogether here, and just set the transient value, which will update it if it already exists.

src/includes/traits/Plugin/UpdateUtils.php
```diff
@@ -128,12 +128,18 @@ public function preSiteTransientUpdatePlugins($transient)
...
-        if (empty($_r[GLOBAL_NS.'_update_pro_version']) || empty($_r[GLOBAL_NS.'_update_pro_zip'])) {
+        if (empty($_r[GLOBAL_NS.'_update_pro_version'])) {
             return $transient; // Nothing to do here.
         }
+
         $update_pro_version = (string) $_r[GLOBAL_NS.'_update_pro_version'];
-        $update_pro_zip     = base64_decode((string) $_r[GLOBAL_NS.'_update_pro_zip'], true);
-        // @TODO Encrypt/decrypt to avoid mod_security issues. Base64 is not enough.
+
+        if (false === ($update_pro_zip = get_site_transient(GLOBAL_NS.'_update_pro_zip_'.$update_pro_version))) {
```
jaswsinc: You probably want to fail here if `$update_pro_zip` is empty in any way, not just if it is is `false` explicitly.

```php
if(!($update_pro_zip = get_site_transient(GLOBAL_NS.'_update_pro_zip_'.$update_pro_version))) {
    return $transient;
}
```

jaswsinc:
> @jaswsinc As noted in websharks/comet-cache#416 (comment), this currently does not work as expected. I get "The plugin is at the latest version." Not sure why.

Looks good to me. I'm not seeing a bug that would cause this, but given the details you've gathered thus far I think it this line would be a good place to look for problems. https://github.com/websharks/comet-cache-pro/blob/6382ff6bf591c883e35ebf930236386cc16e7459/src/includes/traits/Plugin/UpdateUtils.php#L138

jaswsinc:
It might help to improve compatibility and future-proof this a bit further if you don't delete the transient once it is read. That way if WordPress happens to call query for this value more than once during a single update process it will continue to work. https://github.com/websharks/comet-cache-pro/blob/6382ff6bf591c883e35ebf930236386cc16e7459/src/includes/traits/Plugin/UpdateUtils.php#L142

Also, CC already is there to help with stale transients.

raamdev:
> but given the details you've gathered thus far I think it this line would be a good place to look for problems

I've already confirmed (using xdebug) that we get past that line; see screenshot above.

jaswsinc:
I think the most likely issue is that WordPress is calling for `get_site_transient( 'update_plugins' )` more than once during a single upgrade process, and because the transient is being deleted, it's no longer there on any subsequent calls that need it in order for the filter to work.

raamdev:
> It might help to improve compatibility and future-proof this a bit further if you don't delete the transient once it is read. That way if WordPress happens to call query for this value more than once during a single update process it will continue to work. https://github.com/websharks/comet-cache-pro/blob/6382ff6bf591c883e35ebf930236386cc16e7459/src/includes/traits/Plugin/UpdateUtils.php#L142

Deleting that line fixed the bug that was producing the "The plugin is at the latest version."  message, though I'm still not sure why.

raamdev:
All working now:

jaswsinc:
Cool. What plugin is that you're using to review transients?

raamdev:
> Cool. What plugin is that you're using to review transients?

https://wordpress.org/plugins/transients-manager/

raamdev:
@jaswsinc Can I merge this now, or do you want to run a few more tests of your own?

jaswsinc:
Looks read to me also. Go for it.

raamdev:
Awesome. Thanks for helping with this! 😄

PR #255, opened 2016-05-27
Improving inline docs for Misc. Auto Clear Options; 

renzms:
see: websharks/comet-cache#781


raamdev:
@renzms The docs now say:
> Relative links for specific URLs should not be used; e.g., `/name-of-post`.

but the very next sentence contradicts this with the example used:

> A wildcard `*` character can also be used when necessary; e.g., `/category/abc-followed-by-*`

renzms:
@raamdev 

> The docs now say:
> Relative links for specific URLs should not be used; e.g., /name-of-post.
> but the very next sentence contradicts this with the example used:
> A wildcard * character can also be used when necessary; e.g., /category/abc-followed-by-*

The first sentence was meant to explain that a specific URL to a single post/page needs to be in full for it to be detected and cleared properly. e.g., `http://domain.com/name-of-post`. I discovered this during testing, if this is supposed to work otherwise as just a relative link, then it is a bug. 

The second sentence refers to a category and wildcard for a catch-all, so supposedly multiple and not necessarily specific URLs. I don't think it contradicts, but maybe needs to be worded differently.

raamdev:
@renzms According to your updated docs ("Relative links for specific URLs should not be used; e.g., `/name-of-post`"), the following would not work:

```
/category/abc-followed-by-*
```

So this sentence that follows would be incorrect:

> A wildcard `*` character can also be used when necessary; e.g., `/category/abc-followed-by-*`

And should instead say something like:

> A wildcard `*` character can also be used when necessary; e.g., `http://example.com/category/abc-followed-by-*`

Also, as Jason noted [here](https://github.com/websharks/comet-cache/issues/781#issuecomment-222915436), we should clarify that a wildcard character can also be used to match all hostnames, e.g., `*/category/abc-followed-by-*`.

renzms:
@raamdev @jaswsinc 

Taking both of your notes into consideration, I've simplified the statement to the inline docs so they don't contradict each other and are easier to understand. Thanks!

screen shot 2016-06-02 at 11 34 13 pm

raamdev:
@renzms Thanks. I think I may have found a bug in the Auto-Clear List of Custom URLs; see https://github.com/websharks/comet-cache/issues/781#issuecomment-223712805.

raamdev:
@renzms Please see Jason's comment here: https://github.com/websharks/comet-cache/issues/781#issuecomment-223719493

So the docs need to be updated accordingly.

Also, I noticed you're missing a space here after 'scheme': `scheme(e.g., to only`

renzms:
@raamdev 

Ready for review thanks!

raamdev:
@renzms Please post screenshots of the changes whenever your PR includes changes to the UI, otherwise whoever is reviewing the PR needs to rebuild the entire plugin and install it just to review the changes.

raamdev:
@renzms I just reviewed this again:



## A few requested changes:

Please strip out these two ambiguous sentences from the description:

>  For **individual specific URLs** to a post/page, if you wish to clear the cache for a specifc scheme (e.g., to only clear the `https://` cache or to only clear the cache for a particular domain that you are using) you can use the full URL including domain name and `http://` or `https://`. Relative links for specific URLs should not be used; e.g., `/name-of-post`.

Then add a blue informational note that goes at the bottom (under the text box) containing the following:

> **Note:** Relative URLs (e.g., `/name-of-post`) should NOT be used. Each entry above should start with `http://` or `https://` and include a fully qualified domain name.

renzms:
screen shot 2016-06-14 at 8 56 58 pm

@raamdev : Updated, thanks!

raamdev:
@renzms See my comment above about where the blue note box should go.

renzms:
screen shot 2016-06-16 at 1 30 07 am

> See my comment above about where the blue note box should go.

@raamdev Fixed! My bad. Thanks!

PR #263, opened 2016-06-21
Link "Cache Stats" button in Admin Bar to the Cache Stats page

renzms:
See websharks/comet-cache#780


renzms:
screen shot 2016-06-22 at 7 08 44 am

@jaswsinc I've changed the Cache Stats link to: `'.esc_attr(add_query_arg(urlencode_deep(['page' => GLOBAL_NS.'-stats']), network_admin_url('/admin.php'))).'`, like the ​*More Info*​ button.

The link is not clickable as is. But if you right click it and `open in new tab` it goes to the link.

Raam noted I may need to work on the JS too, any pointers on how to fix that?

src/includes/traits/Plugin/AdminBarUtils.php
```diff
@@ -231,7 +231,7 @@ public function adminBarMenu(\WP_Admin_Bar &$wp_admin_bar)
                     'id'     => GLOBAL_NS.'-stats',
 
                     'title' => __('Cache Stats', SLUG_TD),
-                    'href'  => '#',
+                    'href'  => ''.esc_attr(add_query_arg(urlencode_deep(['page' => GLOBAL_NS.'-stats']), network_admin_url('/admin.php'))).'',
```
jaswsinc: @renzms You can make this a little cleaner by removing the double quotes; i.e., `=> ''.esc_attr(` on the beginning and end of this string would only be needed if there was something inside them; e.g., `=> 'something'.esc_attr(`. In this case there is not, so it can just be `=> esc_attr( ...`.

jaswsinc:
> Raam noted I may need to work on the JS too, any pointers on how to fix that?

Try commenting out [this line](https://github.com/websharks/comet-cache-pro/blob/160521/src/client-s/js/admin-bar.js#L58) and then minify the JS file and give it a shot.

src/includes/traits/Plugin/AdminBarUtils.php
renzms: > @renzms You can make this a little cleaner by removing the two single-quotes; i.e., => ''.esc_attr( on the beginning and end of this string would only be needed if there was something inside them; e.g., => 'something'.esc_attr(. In this case there is not, so it can just be => esc_attr( ....

Thanks for tip!

renzms:
> Try commenting out this line and then minify the JS file and give it a shot.

@jaswsinc That worked! Thanks for all the tips!

@raamdev Tested and ready for review!

raamdev:
@renzms Tested and merged. Thanks!

PR #264, opened 2016-06-24
Auto-Cache Engine: Use cURL as a fallback when PHP allow_fopen_url=0

raamdev:
See websharks/comet-cache#440

### DIFF
--- a/src/includes/classes/AutoCache.php
+++ b/src/includes/classes/AutoCache.php
@@ -286,7 +286,7 @@ class AutoCache extends AbsBase
             unset($_sitemapindex_urls, $_urlset_urls);
         }
 
-        if (!$allow_url_fopen) {
+        if (!$allow_url_fopen || $this->applyWpFilters(GLOBAL_NS.'_auto_cache_use_curl', false)) {
             unlink($sitemap); // Delete temp file downloaded by cURL
         }
 
--- a/src/includes/classes/AutoCache.php
+++ b/src/includes/classes/AutoCache.php
@@ -246,6 +246,9 @@ class AutoCache extends AbsBase
      * @throws \Exception If `$sitemap` is NOT actually a sitemap.
      *
      * @return array URLs from an XML sitemap deeply.
+     *
+     * @note The XMLReader class reads the input XML from a stream, but if the PHP config has `allow_fopen_url=0`
+     *       we need to download the XML Sitemap and read it from a local file. We use the WP HTTP API as a fallback.
      */
     protected function getSitemapUrlsDeep($sitemap, $___recursive = false)
     {
@@ -259,8 +262,8 @@ class AutoCache extends AbsBase
         if (!$this->plugin->autoCacheCheckXmlSitemap($sitemap, $___recursive, $this->is_child_blog)) {
             goto finale; // Nothing we can do.
         }
-        if (!$allow_url_fopen || $this->applyWpFilters(GLOBAL_NS.'_auto_cache_use_curl', false)) { // Try cURL instead
-            $sitemap = $this->plugin->autoCacheGetSitemapViaCurl($sitemap);
+        if (!$allow_url_fopen || $this->applyWpFilters(GLOBAL_NS.'_auto_cache_sitemap_force_wp_http_api', false)) { // Try the WP HTTP API instead.
+            $sitemap = $this->plugin->autoCacheWpRemoteGetXmlSitemap($sitemap);
         }
 
         if ($xml_reader->open($sitemap)) {
@@ -286,7 +289,7 @@ class AutoCache extends AbsBase
             unset($_sitemapindex_urls, $_urlset_urls);
         }
 
-        if (!$allow_url_fopen || $this->applyWpFilters(GLOBAL_NS.'_auto_cache_use_curl', false)) {
+        if (!$allow_url_fopen || $this->applyWpFilters(GLOBAL_NS.'_auto_cache_sitemap_force_wp_http_api', false)) {
             unlink($sitemap); // Delete temp file downloaded by cURL
         }
 
--- a/src/includes/classes/AutoCache.php
+++ b/src/includes/classes/AutoCache.php
@@ -249,15 +249,18 @@ class AutoCache extends AbsBase
      */
     protected function getSitemapUrlsDeep($sitemap, $___recursive = false)
     {
-        $urls       = [];
-        $xml_reader = new \XMLReader();
-        $failure    = ''; // Initialize.
+        $urls            = [];
+        $xml_reader      = new \XMLReader();
+        $allow_url_fopen = filter_var(ini_get('allow_url_fopen'), FILTER_VALIDATE_BOOLEAN);
 
         if (!($sitemap = trim((string) $sitemap))) {
             goto finale; // Nothing we can do.
         }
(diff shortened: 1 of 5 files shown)


raamdev:
@jaswsinc I'd love a second set of eyes on this. :-) I ran a couple of tests:

- PHP 5.5 + `allow_fopen_url=0` (falls back to cURL)
- PHP 7 + `allow_fopen_url=1` (works as expected)

In both cases the Auto-Cache Engine works as expected.

**Note:** You can force the use of cURL via a filter in an MU-Plugin:

```php
add_filter('comet_cache_auto_cache_use_curl', function() { return true; });
```

raamdev:
Also note that it's possible to add additional cURL options via a filter, which may be necessary during testing locally due to SSL verification issues:

```php
add_filter('comet_cache_auto_cache_curl_options', function() {
     return array(CURLOPT_SSL_VERIFYHOST => false, CURLOPT_SSL_VERIFYPEER => false);
});
```

src/includes/traits/Plugin/AutoCacheUtils.php
```diff
@@ -170,5 +170,49 @@ public function autoCacheCheckXmlSitemap($sitemap, $is_nested_sitemap = false, $
...
+     *
+     * @return string Path to downloaded XML Sitemap file
+     */
+    public function autoCacheGetSitemapViaCurl($url)
+    {
+        if (!($tmp_dir = $this->getTmpDir())) {
+            throw new \Exception(__('No writable tmp directory.', SLUG_TD));
+        }
+
+        $tmp_file = $tmp_dir.'/'.$this->addTmpSuffix($tmp_dir).'.xml';
```
jaswsinc: @raamdev Lookin' great!

On this line _(this is an inline code comment)_... the `addTmpSuffix()` call is probably not what you want because that function is designed to add a tmp suffix to an existing directory for the purpose of working on a temporary copy of an existing directory; e.g., renaming the cache directory and working on a temporary copy while the cache is locked. So `addTmpSuffix()` will return the full absolute path, but with a `-tmp` suffix.

In the use case you have here, the full path to the `$tmp_file` is probably not what you expect it to be.

You might be better off using: https://developer.wordpress.org/reference/functions/wp_tempnam/

src/includes/classes/AutoCache.php
```diff
@@ -282,6 +285,11 @@ protected function getSitemapUrlsDeep($sitemap, $___recursive = false)
             }
             unset($_sitemapindex_urls, $_urlset_urls);
         }
+
+        if (!$allow_url_fopen) {
```
jaswsinc: Does this conditional check that deletes the temporary file also consider the case of the filter being used?

raamdev: Yes, I updated that in https://github.com/websharks/comet-cache-pro/pull/264/commits/6219b3515a37f7f214b25d86cec918b8f9954386; so you're looking at an old version. :-)

src/includes/traits/Plugin/AutoCacheUtils.php
raamdev: Hmm, the PHPDoc Block for `addTmpSuffix()` should be updated in that case. It currently implies that it can be used for a file path:

```php
/**
 * Adds a tmp name suffix to a directory/file path.
 * ...
*/
```

Looking over the source for `addTmpSuffix()` tells me that it works just as I would expect, so I'm not sure why you say it's not what I'd expect.

src/includes/traits/Plugin/AutoCacheUtils.php
```diff
@@ -170,5 +170,49 @@ public function autoCacheCheckXmlSitemap($sitemap, $is_nested_sitemap = false, $
...
+        }
+
+        $ch = curl_init();
+
+        curl_setopt($ch, CURLOPT_URL, $url);
+        curl_setopt($ch, CURLOPT_FILE, $fp);
+        curl_setopt($ch, CURLOPT_HEADER, 0);
+        curl_setopt($ch, CURLOPT_TIMEOUT, 10);
+        curl_setopt($ch, CURLOPT_FAILONERROR, true);
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
```
jaswsinc: In this line, setting `CURLOPT_FOLLOWLOCATION` to `true` on an installation of PHP that has `safe_mode` on, or has an `open_basedir` is likely to result in an error due to a security precaution in PHP. That's bitten me in the past, so just wanted to let you know about this. See: https://github.com/WordPress/WordPress/blob/602b51a209e39462e45003378df2bf89d008aa72/wp-includes/class-wp-http-curl.php#L147

Or, you might consider converting this to `wp_remote_get()`, which will use cURL first and then automatically fallback on the `Http` extension if it exists. It also deals with quirks like the location redirect flag.

So something like...

```php
$response = wp_remote_get( ... );
if ($response && !is_wp_error($response)) {
    file_put_contents($tmp_file, $response['body']);
}
```

src/includes/traits/Plugin/AutoCacheUtils.php
raamdev: Hmm, I see what you mean. `$tmp_dir.'/'.$this->addTmpSuffix($tmp_dir)`—I'm adding `$tmp_dir` to the path twice. 

I'll refactor that to use `wp_tempnam()` instead. Thanks!

jaswsinc:
Regarding `addTmpSuffix()`, here's a look at why I say it might not be what you expect.

```php
$tmp_dir = '/tmp/dir';
$tmp_dir = $this->addTmpSuffix($tmp_dir);
# Gives you: /tmp/dir-xxxxx-tmp
```

So in this use case:

```php
$tmp_dir = '/tmp/dir';
$tmp_file = $tmp_dir.'/'.$this->addTmpSuffix($tmp_dir).'.xml';
# Gives you: /tmp/dir/tmp/dir-xxxxx-tmp
```

raamdev:
> # Gives you: /tmp/dir/tmp/dir-xxxxx-tmp

Hmm, odd. I wonder why it worked at all then... I saw the temp files get written to `/tmp/4b3403665fea6 .xml` as I was expecting.

raamdev:
@jaswsinc Thanks for the suggestions. :-) See https://github.com/websharks/comet-cache-pro/pull/264/commits/3569386438586af519e85d61364251a3e631a201. I tested this and it's working nicely. 

I'm not sure why it didn't occur to me to use `wp_remote_get()`—whenever I hear "cURL" I automatically think of the way I used it when I was first learning PHP, i.e., outside of WordPress. ;)

raamdev:
@jaswsinc Woops, forgot to take out the call to `fopen()`. Fixed in https://github.com/websharks/comet-cache-pro/pull/264/commits/45a15caff161e1f54b34b231e804d5e1b47fd300.

raamdev:
A few changes:

The filter to force downloading the XML Sitemap using the WP HTTP API has been renamed from `comet_cache_auto_cache_use_curl` to `comet_cache_auto_cache_sitemap_force_wp_http_api`:

```php
add_filter('comet_cache_auto_cache_sitemap_force_wp_http_api', function() { return true; });
```

I also renamed `autoCacheGetSitemapViaCurl()` to `autoCacheWpRemoteGetXmlSitemap()`.

jaswsinc:
Awesome! 💯

PR #265, opened 2016-06-24
Attempt to avoid a race condition in checkCronSetup()

raamdev:
See https://github.com/websharks/comet-cache/issues/653#issuecomment-212282244

### DIFF
--- a/src/includes/traits/Plugin/CronUtils.php
+++ b/src/includes/traits/Plugin/CronUtils.php
@@ -34,7 +34,7 @@ trait CronUtils
      */
     public function checkCronSetup()
     {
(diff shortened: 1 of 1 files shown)


src/includes/traits/Plugin/CronUtils.php
```diff
@@ -34,7 +34,7 @@ public function extendCronSchedules($schedules)
      */
     public function checkCronSetup()
     {
-        if ($this->options['crons_setup'] < 1439005906
+        if ((!get_transient('doing_cron') && $this->options['crons_setup'] < 1439005906)
```
jaswsinc: Neat.

PR #270, opened 2016-07-14
Improve formatting of HTML notes; See: websharks/comet-cache#790

kristineds:
Improve formatting of HTML notes; 

See: websharks/comet-cache#790
---

_Note: Work in progress_


jaswsinc:
@kristineds The way I suggest going about this would be to build each note as a line of its own, and put each of those lines into an an associative array with keys as headings.

```php
$notes = []; // Initialize array.
$notes['File Path'] = 'blah blah blah';
$notes['Page URL']  = 'blah blah blah';
```

With that in place, you can then come back and iterate the array in interesting ways; e.g., we are hoping to build a tabulated list. Something like this.

```php
$longest_heading_size = 0; // Initialize.

// Find the longest heading size.
foreach ($notes as $_heading => $_value) {
    $longest_heading_size = max($longest_heading_size, mb_strlen($_heading));
} unset($_heading, $_value); // Housekeeping.

// Build a tabulated list.
foreach ($notes as $_heading => $_value) {
    echo '<!-- '.htmlspecialchars(str_pad($_heading, $longest_heading_size).': '.$_value).' -->';
} unset($_heading, $_value); // Housekeeping.
```

---

Another advantage of each line being in an associative array, is that it makes it easier for a specific line to be excluded or included based on a configuration value.

e.g., `unset($notes['File Path']);`

kristineds:
@jaswsinc I've added the arrays but I don't think I'm doing it right. 😞  Would you mind looking at it when you have a chance? https://github.com/websharks/comet-cache-pro/pull/270/commits/01f4c56c874d9b15f39c720f7ad8697dbd257692

src/includes/traits/Ac/ObUtils.php
```diff
@@ -348,14 +348,32 @@ public function outputBufferCallbackHandler($buffer, $phase)
...
-            $cache .= "\n".htmlspecialchars(sprintf(__('Generated via: \t10s %6$s', SLUG_TD), NAME, $this->is_404 ? '404 [error document]' : $this->salt_location, (IS_PRO && COMET_CACHE_WHEN_LOGGED_IN && $this->user_token ? '; '.sprintf(__('user token: %1$s', SLUG_TD), $this->user_token) : ''), $total_time, date('M jS, Y @ g:i a T'), $via));
-            $cache .= "\n".' -->';
+
+            $cache .= $notes;
+
+            $notes = []; // Initialize array.
+            $notes['Cache File Information and Statistics'] = "\n".'++++++++++++++++'.htmlspecialchars(sprintf(__(' ', SLUG_TD), NAME, str_replace(WP_CONTENT_DIR, '', $this->is_404 ? $this->cache_file_404 : $this->cache_file)));
+            $notes['File Path'] = "\n".htmlspecialchars(sprintf(__('%2$s', SLUG_TD), NAME, str_replace(WP_CONTENT_DIR, '', $this->is_404 ? $this->cache_file_404 : $this->cache_file)));
```
jaswsinc: You can remove the `\n` line breaks from each item in the array, because these can be applied later when the array is being iterated and each of the lines are brought together as a tabulation. Also, the call to `htmlspecialchars()` in each of those array items can be removed also. That's also something that we can do in the array iteration later as it is brought together.

src/includes/traits/Ac/ObUtils.php
```diff
@@ -348,14 +348,32 @@ public function outputBufferCallbackHandler($buffer, $phase)
...
+            $notes['Date generated'] = "\n".htmlspecialchars(sprintf(__('%5$s', SLUG_TD), NAME, $this->is_404 ? '404 [error document]' : $this->salt_location, (IS_PRO && COMET_CACHE_WHEN_LOGGED_IN && $this->user_token ? '; '.sprintf(__('user token: %1$s', SLUG_TD), $this->user_token) : ''), $total_time, date('M jS, Y @ g:i a T')));
+            $notes['Expiration date'] = "\n".htmlspecialchars(sprintf(__('%2$s (based on your configured expiration time)', SLUG_TD), NAME, date('M jS, Y @ g:i a T', strtotime('+'.COMET_CACHE_MAX_AGE))));
+            $notes['Time to generate this cache file:'] = "\n".htmlspecialchars(sprintf(__('%4$s seconds', SLUG_TD), NAME, $this->is_404 ? '404 [error document]' : $this->salt_location, (IS_PRO && COMET_CACHE_WHEN_LOGGED_IN && $this->user_token ? '; '.sprintf(__('user token: %1$s', SLUG_TD), $this->user_token) : ''), $total_time, date('M jS, Y @ g:i a T')));
```
jaswsinc: You can exclude the `:` from the heading in this line. We can add this when each item is formatted below.

src/includes/traits/Ac/ObUtils.php
```diff
@@ -348,14 +348,32 @@ public function outputBufferCallbackHandler($buffer, $phase)
...
+            $notes['Generated via'] = "\n".htmlspecialchars(sprintf(__('%6$s', SLUG_TD), NAME, $this->is_404 ? '404 [error document]' : $this->salt_location, (IS_PRO && COMET_CACHE_WHEN_LOGGED_IN && $this->user_token ? '; '.sprintf(__('user token: %1$s', SLUG_TD), $this->user_token) : ''), $total_time, date('M jS, Y @ g:i a T'), $via));
```
jaswsinc: I suggest using Title Caps in each of the headings; e.g., `Generated Via`.

src/includes/traits/Ac/ObUtils.php
```diff
@@ -348,14 +348,32 @@ public function outputBufferCallbackHandler($buffer, $phase)
...
+
+             $cache .= "\n".' -->';
+
+            $longest_heading_size = 0; // Initialize.
+
+            // Find the longest heading size.
+            foreach ($notes as $_heading => $_value) {
```
jaswsinc: Once you get this working (or before, whichever), I'd suggest moving these iterations and the formatting that they do to a new utility method somewhere else in the codebase; i.e., make that a function that accepts an array of items that it will tabulate. That way it be reused if we want to.

kristineds:
@jaswsinc Thank you the feedback! I'm sending an _in progress_ PR for this while I'm still trying to figure out how to create that utility method. If you could give a hint 😉  or some kind of existing example that I could look at, that will be very much appreciate. cc @raamdev 

**UPDATED CHANGES:** 

- [ ] Remove the `\n` line breaks from each item in the array,
- [ ] Remove call to `htmlspecialchars()` in each of those array items
- [ ] Exclude the `:` from the heading
- [ ] Use Title Caps in each of the headings; e.g., `Generated Via`.

kristineds:
@raamdev I updated the code for the `$notes` array. Could you point me in the right direction on where exactly I should be adding the `buildTabulatedList()`? 😃   cc @jaswsinc . https://github.com/websharks/comet-cache-pro/pull/270/commits/492b422f49ff3b1e01ea4db2552575d3eb6d502d

raamdev:
Noting that we should take into consideration https://github.com/websharks/comet-cache/issues/818 here, where the data generated for the HTML Notes might want to be used for something other than outputting as HTML Notes.

raamdev:
@kristineds writes...
> Could you point me in the right direction on where exactly I should be adding the buildTabulatedList()?

@jaswsinc Considering that we may want to expand this into providing access to this info in various ways, what do you think about a new `CacheInfoUtils.php` trait in `traits/Shared/`?

jaswsinc:
@raamdev I think that's a great idea. Taking that a step further you might consider a new class called CacheInfo that will hold each of the lines as properties of that class. Then the class can have methods that output the data in various ways.

Prototype:

```php
class CacheInfo {
    protected $data = [];
    
    public function addData($key, $value) {
        $this->data[$key] = $value;
    }

    public function tabulate() {
        // @TODO
        foreach ($this->data as $_key => $_value) {}
    }
}
```

raamdev:
@jaswsinc Do you think you could help Kristine with this one? ↑

jaswsinc:
Yes. I'll take over on this following a discussion in Slack.
https://websharks.slack.com/archives/comet-cache/p1476163673000096

PR #273, opened 2016-08-28
Clear Cache Button in Admin Toolbar (Lite version)

raamdev:


raamdev:
@renzms Regarding your questions about the Clear Cache button in the Admin Bar: There should be no new options visible in the Lite version and the Clear Cache button should be _only_ a Clear Cache button, not the whole menu of options like we have in the Pro version. In the Lite version, the Clear Cache button in the Admin Bar will behave exactly the way the "Clear Cache" button at the top-right of the Options page behaves in the Lite version, i.e., we're just making it slightly more convenient for site owners with the Lite version by giving them access to the Clear Cache button in the Admin Toolbar.

renzms:
@raamdev 

Addressed the Phing Build errors, however I'm receiving the following fatal error in the WP Dashboard:

```

Parse error: syntax error, unexpected 'break' (T_BREAK) in /home4/wsksdev/public_html/renz/wp-content/plugins/comet-cache/src/includes/traits/Plugin/AdminBarUtils.php on line 33
```

I checked the original [line](https://github.com/websharks/comet-cache-pro/blob/68c5a82d3609592d74532466b34973968625955b/src/includes/traits/Plugin/AdminBarUtils.php#L33): 

```
case 'cache_clear':
```

This is the new line:

```
$showing = $this->options['cache_clear_admin_bar_enable'] && $is_multisite;
```

src/client-s/css/admin-bar.scss
```diff
@@ -1,3 +1,11 @@
+@import '../../vendor/package/bourbon/bourbon-v4.2.3/bourbon';
```
raamdev: @renzms This line and the 3 that follow appear to be duplicated again below `/*![pro strip-from='lite']*/`; I suggest removing the duplicate lines and simply adjusting where `/*![pro strip-from='lite']*/` appears.

src/includes/traits/Plugin/AdminBarUtils.php
```diff
@@ -110,15 +106,31 @@ public function adminBarMenu(\WP_Admin_Bar &$wp_admin_bar)
                 ]
             );
         }
+        /*[/pro]*/
         if ($this->adminBarShowing('cache_clear')) {
+            if (!IS_PRO) {
```
raamdev: @renzms This new `$wp_admin_bar->add_menu()` section is not necessary, as it's already included further below. I suggest removing this `if(!IS_PRO)` statement and simply using [the one below](https://github.com/websharks/comet-cache-pro/blob/2c0ee9b27e56a054629e69109a3c28c4efaf6a29/src/includes/traits/Plugin/AdminBarUtils.php#L187). (You'll need to adjust the `/*[pro strip-from="lite"]*/` markers so that the one below shows up in the Lite version.)

PR #276, opened 2016-09-05
Feature/465

jaswsinc:
Adds support for automatic background updates as best I can tell. I found it very difficult to test this, so if you can share your test workflow with me I'd appreciate it. I haven't dug deep enough yet to be absolutely sure about how this goes down in WordPress core from start to finish.

Having said that, I _was_ able to get this working on at least one occasion via CRON after this patch was applied. It appears this was caused by a lack of the `autoupdate` property in our implementation. See: [this WP core conditional check](https://github.com/WordPress/WordPress/blob/9383bf8f748e65e38a5ed70a29134d6a2b9d9ddc/wp-admin/includes/class-wp-automatic-updater.php#L158).

### DIFF
--- a/src/includes/classes/VsUpgrades.php
+++ b/src/includes/classes/VsUpgrades.php
@@ -43,6 +43,7 @@ class VsUpgrades extends AbsBase
         $this->fromZenCache();
         $this->fromLte160227();
         $this->fromLte160521();
+        $this->fromLte160709();
     }
 
     /**
@@ -257,6 +258,18 @@ class VsUpgrades extends AbsBase
             if ($is_apache) {
                 $this->plugin->enqueueMainNotice(sprintf(__('<strong>New %1$s Feature!</strong> This release of %1$s includes a whole new panel for Apache Performance Tuning. Visit the <a href="%2$s">settings</a> and see the new options in <strong>Comet Cache → Plugin Options → Apache Optimizations</strong>.', SLUG_TD), esc_html(NAME), esc_attr(add_query_arg(urlencode_deep(['page' => GLOBAL_NS]), self_admin_url('/admin.php')))));
             }
+        }
+    }
+
+    /**
+     * Before we replaced the Pro Plugin Updater system with the WordPress Plugin Update system.
+     *
+     * @since $v
+     */
+    protected function fromLte160709()
+    {
+        if (version_compare($this->prev_version, '160709', '<=')) {
+            $this->plugin->dismissMainNotice('new-pro-version-available'); // Dismiss any existing notices like this; upgrade notices are handled by WordPress now.
         }
     }
 }
--- a/src/includes/traits/Plugin/UpdateUtils.php
+++ b/src/includes/traits/Plugin/UpdateUtils.php
@@ -180,6 +180,7 @@ trait UpdateUtils
         if ($latest_version && $latest_package && version_compare($latest_version, VERSION, '>')) {
             $report->response[$plugin_basename] = (object) [
                 'id'          => 0,
+                'autoupdate'  => true,
                 'url'         => $plugin_url,
                 'slug'        => $plugin_slug,
                 'plugin'      => $plugin_basename,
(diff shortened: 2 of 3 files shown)


jaswsinc:
Scratch this. Don't merge yet please. The above is incorrect.

jaswsinc:
_I'll take a closer look at this a bit later._

jaswsinc:
I'll have to look closer, but my feeling is that setting the `autoupdate` property is unnecessary. I think the real underlying problem could have simply been the `E_NOTICE` error being triggered in debug mode, which I fixed here and that's when it started working on my side. I'll need to confirm this however. https://github.com/websharks/comet-cache-pro/pull/276/commits/1662a05eb26428710e89fc404249f64a64137844

jaswsinc:
Setting `autoupdate` is a bad thing from what I can see. That effectively forces an auto-update, which we don't want to do. hehe.

raamdev:
@jaswsinc writes...
> I found it very difficult to test this, so if you can share your test workflow with me I'd appreciate it.

I just used WP Crontrol to edit the `wp_update_plugins` event and changed the "Next run" to `+1 minute` from now, then I let that event run as normal.

raamdev:
Referencing associated GitHub issue: https://github.com/websharks/comet-cache/issues/289

jaswsinc:
This is ready to merge now.
I was able to confirm that automatic background updates are working properly.

@raamdev writes...

> I just used WP Crontrol to edit the wp_update_plugins

I found the CRON job which triggers automatic background updates. It's actually [`wp_version_check`](https://github.com/WordPress/WordPress/blob/a3ffebce3036c69f38e9fd2a798a9e0c13084ea6/wp-includes/update.php#L174) instead of `wp_update_plugins` from what I can tell.

What makes this particularly tricky to test, is that `wp_version_check` will only run once every `60` seconds, even if you keep spawning a new CRON job over and over again. In addition, our own update API that makes this possible will only run, at most, once every hour; i.e., it only checks for a new version, at most, once per hour.

None of this is an issue on a live site, because the CRON job runs every 12 hours, which means that this should do fine in production. When testing though, I found it rather tricky, because you've got to run `wp_version_check` at least 60 seconds apart, and you also need to force the Comet Cache option `last_pro_update_check` to a value of `0`, or wait for up to an hour and let that happen by itself.

raamdev:
It looks like in some scenarios an "Update Available" Dashboard notice might get left behind that contains a link that no longer works:



http://example.com/wp-admin/admin.php?page=comet_cache-pro-updater



We'll need to make sure any such notice gets dismissed, probably as part of a VS upgrade routine?

raamdev:
@jaswsinc writes...
> None of this is an issue on a live site, because the CRON job runs every 12 hours, which means that this should do fine in production. When testing though, I found it rather tricky, because you've got to run wp_version_check at least 60 seconds apart, and you also need to force the Comet Cache option last_pro_update_check to a value of 0, or wait for up to an hour and let that happen by itself.
> 
> The result, when testing, is that you might see some plugins get updated automatically, while Comet Cache Pro is not (yet). It will though, just as soon as the stars align.

I'm still not able to get this working. I've tried waiting 60 seconds between running `wp_version_check` (which I've tried doing by manually running it using WP Crontrol, _and_ by rescheduling it to run on its own +2 minutes from now and then waiting for it to run). I've tried manually resetting `last_pro_update_check` to `0` in the database.

In no scenario am I able to get the Comet Cache Pro plugin to auto-update. It shows an update is available, but the auto-update of Comet Cache Pro never occurs.

raamdev:
@jaswsinc writes...
> just as soon as the stars align.

Yep, that sounds about right. 

After doing a bunch more tests on my local test site, _and_ my internet-accessible test site, I finally saw the auto-update occur, where 3 plugins (including Comet Cache Pro) were auto-updated in the background. However, there were 6 plugins that needed updates, and only 3 were updated. Weird.

Anyway, I'm going to merge this and mark https://github.com/websharks/comet-cache/issues/289 as fixed for the next release.

PR #278, opened 2016-09-06
Add $intersect parameter to updateOptions() for use in VS Upgrades

raamdev:
See websharks/comet-cache#807

### DIFF
--- a/src/includes/traits/Plugin/OptionUtils.php
+++ b/src/includes/traits/Plugin/OptionUtils.php
@@ -11,14 +11,15 @@ trait OptionUtils
      * @since 151002 Improving multisite compat.
      *
      * @param bool  $intersect Discard options not present in $this->default_options
+     * @param bool  $refresh   Force-pull options directly from get_site_option()
      *
      * @return array Plugin options.
      *
      * @note $intersect should be `false` when this method is called via a VS upgrade routine or during inital startup on when upgrading. See https://git.io/viGIK
      */
-    public function getOptions($intersect = true)
+    public function getOptions($intersect = true, $refresh = false)
     {
-        if (!($options = $this->options)) { // Not defined yet?
+        if (!($options = $this->options) || $refresh) { // If not defined yet, or if we're forcing a refresh via get_site_option()
             if (!is_array($options = get_site_option(GLOBAL_NS.'_options'))) {
                 $options = []; // Force array.
             }
@@ -69,7 +70,7 @@ trait OptionUtils
         $this->options = $intersect ? array_intersect_key($this->options, $this->default_options) : $this->options;
         update_site_option(GLOBAL_NS.'_options', $this->options);
 
-        return $this->getOptions();
+        return $this->getOptions($intersect);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


raamdev:
@jaswsinc A quick review here when you get a chance, please.

One thought that I had was that we probably need to call `updateOptions()` once _after_ the VS upgrade routines have run now, so that we filter out any deprecated options, correct? (Now that all the calls to `updateOptions()` in the VS upgrade routines are passing `$intersect = false`, deprecated options won't be cleaned out of the database during those routines.) 

Would it make sense to add the following to [this line](https://github.com/websharks/comet-cache-pro/blob/feature/807/src/includes/traits/Plugin/InstallUtils.php#L53)?

```php
$this->updateOptions();
```

jaswsinc:
Love it! Gotta set this up myself now.

jaswsinc:
> Would it make sense to add the following to this line?

Yes, that sounds like a good to me.

raamdev:
Looks like this fix does not take into consideration upgrades via WP CLI... researching.

raamdev:
@jaswsinc  See my last commit. The reason the bug wasn't fixed even after adding `$intersect` to `updateOptions()` is that `getOptions()` was being called earlier in `Plugin.php` when `$this->options` was filled in, and `getOptions()` was intersecting with the default options, thereby dropping the old options needed by the VS Upgrade routines.

In my last commit, I'm calling `getOptions()` twice in `Plugin.php`, which is not ideal from a performance standpoint, but I'm not seeing an easy way around this because we need to `getOptions()` to compare the version to check if we're upgrading, but we cannot `$intersect = true` until _after_ we figure out if we're upgrading. The downside here is that you end up needing to always call `getOptions()` twice, except when upgrading (ideally, it would be the other way around).

raamdev:
Also: I verified that this branch now works as expected, i.e., the bug we're trying to fix has been fixed and old options are properly migrated over via the VS Upgrade routines.

src/includes/traits/Plugin/OptionUtils.php
```diff
@@ -10,9 +10,13 @@
...
+     * @param bool  $intersect Discard options not present in $this->default_options
+     *
      * @return array Plugin options.
+     *
+     * @note $intersect should be `false` when this method is called via a VS upgrade routine or during inital startup on when upgrading. See https://git.io/viGIK
      */
-    public function getOptions()
+    public function getOptions($intersect = true)
     {
         if (!($options = $this->options)) { // Not defined yet?
```
jaswsinc: @raamdev It might be helpful if you added another new parameter to this method that you might call `$refresh`, and if `$refresh` is true, this check for existing options would be bypassed, thereby pulling the options from the underlying `get_site_option()` call again.

With that in place, you should then be able to remove the double call to `getOptions()` in the `setup()` routine, and instead, in the `checkVersion()` method, if there _is_ a new version, before the VS upgrades run, make another call to refresh the options array, with `$intersect = false` and `$refresh = true`.

jaswsinc: Let me know if that's not clear. I can open a new fork and try to explain in code.

raamdev:
@jaswsinc See my last commit, following your advice to add a `$refresh` param. I re-tested and everything is working great.

jaswsinc:
@raamdev This looks wonderful to me.

PR #280, opened 2016-09-15
Display Dashboard notice when there's a Pro Update error

raamdev:

### DIFF
--- a/src/includes/classes/Actions.php
+++ b/src/includes/classes/Actions.php
@@ -563,6 +563,7 @@ class Actions extends AbsBase
         if ($this->plugin->options['enable'] && !$args['cdn_enable']) {
             $args['htaccess_access_control_allow_origin'] = 0;
         }
+        $this->plugin->dismissMainNotice('pro_update_error'); // Clear any previous error notice in case the issue has been fixed.
         /*[/pro]*/
         $args = $this->plugin->trimDeep(stripslashes_deep((array) $args));
         $this->plugin->updateOptions($args); // Save/update options.
--- a/src/includes/traits/Plugin/UpdateUtils.php
+++ b/src/includes/traits/Plugin/UpdateUtils.php
@@ -116,9 +116,16 @@ trait UpdateUtils
         $product_api_response = wp_remote_post($product_api_url, ['body' => $product_api_input_vars]);
         $product_api_response = json_decode(wp_remote_retrieve_body($product_api_response));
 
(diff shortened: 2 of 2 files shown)


raamdev:
@jaswsinc Review here when you get a chance, please.

src/includes/classes/Actions.php
```diff
@@ -563,6 +563,7 @@ protected function saveOptions($args)
         if ($this->plugin->options['enable'] && !$args['cdn_enable']) {
             $args['htaccess_access_control_allow_origin'] = 0;
         }
+        $this->plugin->dismissMainNotice('pro_update_error'); // Clear any previous error notice in case the issue has been fixed.
```
jaswsinc: This looks good to me.

src/includes/traits/Plugin/UpdateUtils.php
```diff
@@ -116,9 +116,16 @@ protected function maybeCheckLatestProPackage()
         $product_api_response = wp_remote_post($product_api_url, ['body' => $product_api_input_vars]);
         $product_api_response = json_decode(wp_remote_retrieve_body($product_api_response));
 
+        $this->dismissMainNotice('pro_update_error'); // Clear any previous error notice in case a previous issue has been fixed.
```
jaswsinc: This looks good also.

src/includes/traits/Plugin/UpdateUtils.php
```diff
@@ -116,9 +116,16 @@ protected function maybeCheckLatestProPackage()
...
+
+        if (!empty($product_api_response->error)) {
```
jaswsinc: No problems here either.

raamdev: @jaswsinc One thing that concerned me here was the scenario we fixed by implementing the proxy fallback, where an older version of cURL might be unable to connect to cometcache.com. I did a little research but couldn't figure out if in that scenario `$product_api_response->error` would be set. If it would, then we need to figure out how to work around that here, since in that case Comet Cache would be incorrectly reporting a username/password error and would never even try the proxy server below...

jaswsinc: If the connection fails then `->error` will not be set. That particular property comes from the cometcache.com server, so it the connection fails, `$product_api_response` will be empty.

raamdev: @jaswsinc Perfect. Thanks!

PR #293, opened 2016-11-08
PR: feature/639

jaswsinc:
See: websharks/comet-cache#639


raamdev:
Fixed the opening comment to point to the correct GitHub issue (websharks/comet-cache#639 not websharks/comet-cache-pro#639).

jaswsinc:
Referencing new redirect:
https://cometcache.com/r/google-analytics-variables/

jaswsinc:
Default value for new option: `utm_*`

jaswsinc:
New constant: `COMET_CACHE_IGNORE_GET_REQUEST_VARS` (regex).

lkraav:
@jaswsinc thoughts about `COMET_CACHE_IGNORE_GET_REQUEST_VARS` it looks like we currently have a blacklist. One of my scenarios would actually benefit from whitelisting instead.

`COMET_CACHE_IGNORE_GET_REQUEST_VARS === *`

`COMET_CACHE_ALLOW_GET_REQUEST_VARS === [ 'date', 'location' ]` etc.

In this scenario, the default is to route cached requests as if no query parameters even exist.

Only if ALLOW list exists, those specific parameters would create new cache entries.

Your thoughts?

jaswsinc:
> COMET_CACHE_ALLOW_GET_REQUEST_VARS

@lkraav Yes, that does sound like a great idea to me also.



> This would lower DoS surface quite a bit. Your thoughts?

Agree. That would help to eliminate the possibility that an attacker could hit URLs with query strings just to avoid your cache engine and eat up more processing time.

@lkraav Would you mind collecting your thoughts and mine and then opening a new issue for this please? @raamdev Can add this to a future release cycle for us.

PR #294, opened 2016-11-10
PR: feature/793

jaswsinc:
See: websharks/comet-cache#793


jaswsinc:
Regarding point 2 in https://github.com/websharks/comet-cache/issues/793

For Apache, we already have this in place for the cache directory.

```
<IfModule authz_core_module>
	Require all denied
</IfModule>
<IfModule !authz_core_module>
	deny from all
</IfModule>
```

For Nginx, site owners should take it upon themselves to secure cache directories as a part of the server setup whenever you are running WordPress with a caching plugin.

I'm adding the following this to KBA:
https://cometcache.com/kb-article/recommended-nginx-configuration-for-comet-cache/

```nginx
# Do not allow public access to private cache directory.
if ($uri ~* /wp\-content/cache/comet\-cache/cache(?:/|$)) {
    return 403;
}
```

```nginx
# Do not allow public access to private cache directory.
if ($uri ~* /wp\-content/cache/comet\-cache/htmlc/private(?:/|$)) {
    return 403;
}
```

jaswsinc:
- Adding Nonce option.
- Removing Admin Bar option.

PR #296, opened 2016-12-07
AMP compat. See: https://github.com/websharks/comet-cache/issues/695

jaswsinc:
AMP compat. See: https://github.com/websharks/comet-cache/issues/695


readme.txt
```diff
@@ -1,8 +1,8 @@
 === Comet Cache ===
 
-Stable tag: 161119
+Stable tag: 161207
 Requires at least: 4.2
-Tested up to: 4.7
+Tested up to: 4.7-alpha
```
raamdev: @jaswsinc This should not be `4.7-alpha`, since `4.7` has been released. That's the reason this was changed. See https://github.com/websharks/comet-cache-pro/commit/34c3d1f4698b9432425845f501557ea26ce2a079.

jaswsinc: Good catch. Hmm. I'm not sure how this got changed back to `-alpha` then. I don't remember changing that myself, but I guess I must have.

jaswsinc: I will update this in other work and make sure it's as `4.7` in the dev branch.

PR #297, opened 2016-12-07
PR: Automatic Background Updates

jaswsinc:
See: websharks/comet-cache#827


src/includes/classes/MenuPageOptions.php
```diff
@@ -234,10 +234,19 @@ public function __construct()
             echo '      <h3>'.sprintf(__('License Key', SLUG_TD), esc_html(NAME)).'</h3>'."\n";
             echo '      <p><input type="password" name="'.esc_attr(GLOBAL_NS).'[saveOptions][pro_update_password]" value="'.esc_attr($this->plugin->options['pro_update_password']).'" autocomplete="new-password" /></p>'."\n";
 
+            if ((!defined('DISALLOW_FILE_MODS') || !DISALLOW_FILE_MODS) && !apply_filters('automatic_updater_disabled', defined('AUTOMATIC_UPDATER_DISABLED') && AUTOMATIC_UPDATER_DISABLED)) {
```
raamdev: @jaswsinc Shouldn't `&& AUTOMATIC_UPDATER_DISABLED` be `&& !AUTOMATIC_UPDATER_DISABLED`?

jaswsinc: A little tricky there, but I believe that's right like it is. `defined('AUTOMATIC_UPDATER_DISABLED') && AUTOMATIC_UPDATER_DISABLED)` establish the default value for the updater being disabled, which is then filtered, and we check if it's not being filtered; i.e., `!apply_filters(...`

raamdev: I see, you're right. I missed the fact that `&& AUTOMATIC_UPDATER_DISABLED` was still inside `apply_filters()`. Looks right to me now. Thanks!

PR #307, opened 2016-12-23
[PR] Updating via IWP temporarily broke sites

jaswsinc:
See: websharks/comet-cache#862


jaswsinc:
## Summary of Changes

- New constant in `advanced-cache.php`: `COMET_CACHE_AC_FILE_VERSION`. This allows the `AdvancedCache{}` class to stop when the `advanced-cache.php` file was built by a version of the plugin that is not a match to the currently installed plugin version. Or, if that new constant is not defined at all, it will also stop; e.g., upgrading from a version prior to us adding that constant.

- I also moved `checkVersion()` from `admin_init` to the `init` hook.

---

The effect is that upgrading Comet Cache (lite or pro) from a remote management tool makes the `advanced-cache.php` file stale. In such a scenario, the first HTTP request that occurs after an upgrade via a remote management tool results in Comet Cache being disabled automatically — for that first request, when `advanced-cache.php` is stale.

At which point the plugin version is checked on `init` now, instead of on `admin_init`. The plugin upgrade routines run and the `advanced-cache.php` is rebuilt during this process. Therefore, the second request after upgrading via a remote management  automatically begins caching again, with a clean slate; i.e., right after the full Comet Cache upgrade process completes.

PR #311, opened 2017-01-27
[PR] Cache for date-based archive views not cleared when post is updated

jaswsinc:
See: websharks/comet-cache#724


jaswsinc:
---

PR #314, opened 2017-04-25
[PR] Feature Request: Add support for Memcached and Amazon ElastiCache

jaswrks:
See: websharks/comet-cache#47


jaswrks:
### Testing Memcached on wpsharks.net Test Sites
Enter the following as your Memcached server: `127.0.0.1:11211:1`

jaswrks:
@raamdev This is ready to merge/review when you have time. You can use the details above to test this on your wpsharks.net test site if you'd like.

jaswrks:
### Worth Noting

- The menu page options class was broken apart as a part of the work in this PR. See: https://github.com/websharks/comet-cache-pro/tree/feature/47/src/includes/classes/MenuPage/Options

- A new dependency was added to complete work in this PR: `"ramsey/uuid": "3.*"`, see: https://github.com/websharks/comet-cache-pro/blob/feature/47/composer.json

raamdev:
Should I be seeing _slower_ speeds when loading from RAM? I was getting a consistent loading speed of `0.04357 seconds` when using RAM / Memcached during testing and when I disabled RAM / Memcached and switched back to disk-based caching the loading speed dropped, quite considerably and consistently, to `0.00557 seconds`.

Is there anything else that you recommend doing as far as testing (besides simply checking if it says `via-memory` in the HTML Notes)?

----

The placeholder text is missing the colons:



----
In the description, the following text:

> a Memcached server with a minimum of 512MB of RAM available and generally no more than 5GB are necessary

Should be...

> a Memcached server with a minimum of 512MB of RAM available and generally no more than 5GB **is** necessary

----

Otherwise, this looks fantastic! I'm so excited to see this much-anticipated feature finally becoming a reality. :-)

jaswrks:
## User-Specific Caching

We may need to disable Memcached automatically for logged-in user caching, as this tends to cause a few problems. One being that it somewhat unnecessarily eats RAM for user-specific cache entries, which are already quite transient. Another being that it results in the entire RAM-based cache being wiped out (for all users, logged-in or not) whenever any single user performs an action that requires the cache to be cleared in some way.

Limitations in the simple approach taken by Memcached (in terms of how we can clear the cache) are accentuated by the need to manage a set of user-specific cache entries in RAM, which requires that we clear those separately, and that's nearly impossible to do effectively.

jaswrks:
@raamdev writes...

> Should I be seeing slower speeds when loading from RAM? I was getting a consistent loading speed of 0.04357 seconds when using RAM / Memcached during testing and when I disabled RAM / Memcached and switched back to disk-based caching the loading speed dropped, quite considerably and consistently, to 0.00557 seconds.

No, that seems like quite a spread to me. You may see some fluctuation, and it's even possible that memory could be slightly slower in some cases, but it should generally be the same speed or faster on those DigitalOcean droplets. Please let me know if that continues. Also try the latest version after the tweaks I made earlier.

Also, here are some things worth noting when running tests.

- Our wpsharks.net test sites run SSDs, so the difference between RAM and filesystem delivery is not nearly as much as you'd see when comparing RAM to an optical drive.

- On a test site that has little or no concurrent traffic, especially one that's running an SSD, we may need to create artificial traffic spikes to see the benefit of a RAM-based delivery.

- Memcached gets slightly faster the longer it runs. One reason being that PHP-FPM runs multiple processes via FastCGI, and each of those acquires a persistent connection to the Memcached server. Therefore, it's hit or miss in the very beginning while those persistent connections are being established across multiple FastCGI processes on the backend. When you first start testing, you may see some wild fluctuations early-on. It shouldn't be the difference between `0.04357` and `0.00557`, but I have seen new connections causing one hit to the RAM cache taking `0.03` while the next takes `0.016`. The longer time represents a new connection to Memcached instead of being able to reuse an existing persistent connection that's already been opened.

---

You can also interact with Memcached from the command-line.

```bash
echo stats | nc 127.0.0.1 11211;
```

Or connect and speak with Memcached like this.

```bash
nc 127.0.0.1 11211
```

```bash
stats items
```

jaswrks:
> The placeholder text is missing the colons:
> 5GB is necessary

Thanks. I fixed those.

jaswrks:
> Is there anything else that you recommend doing as far as testing (besides simply checking if it says via-memory in the HTML Notes)?

## Other Items We Need to Test

- The Memcached library this uses comes from the WebSharks Core and that's only ever been tested on PHP 7. I adjusted it for compatibility with PHP 5.4 whenever I ported it to Comet Cache, but we still need to find a way of running tests against this on older versions of PHP that are running older versions of the Memcached extension too. Just to see if there are any incompatibilities we can work around or need to mention in the documentation. An example might be a case where one of the `\Memcached::CONSTANTS` did not exist in older versions of PHP. I'm not aware of that being the case, but it's possible.

- Need to test with multiple Memcached servers, not just one, and then we need to hit multiple pages of a site when there are multiple Memcached servers configured. The hashing algorithm used by Memcached will automatically spread the cache out across all of the servers you configure, which is done by sharding on a hash of the cache key. In short, the cache for one page will be in server A, and another will be on server B. So testing with multiple servers and then hitting multiple pages should prove that's working or not working.

- Need to test this on a multisite network. I have not tested that at all yet.

- Need to test clearing the cache, because clearing the cache results in the entire RAM-based cache being wiped away each time. For example, if you update a Post, that clears the entire RAM-based cache for all cache entries. The filesystem is cleared intelligently, memory is not. It's just wiped.

- We need to consider [this point I raised above](https://github.com/websharks/comet-cache-pro/pull/314#issuecomment-297611400).

jaswrks:
- Also need to test against AWS ElastiCache. I'm not aware of that creating any new quirks, but it's been about 2 or 3 years since I used ElastiCache. Things may have changed since then.

raamdev:
> Should I be seeing slower speeds when loading from RAM?

I re-tested this speed issue with a new build after your most recent changes and I'm no longer seeing a problem. The loading speed is consistently around `0.00557`, for both memory and disk (which would make sense in this case given that the server is using an SSD disk). 

> Other Items We Need to Test

Which of those tests do you feel should be done before this goes to RC?

jaswrks:
> Which of those tests do you feel should be done before this goes to RC?

I could use some guidance on [the User-Specific RAM-based cache](https://github.com/websharks/comet-cache-pro/pull/314#issuecomment-297611400). That's what I think needs internal testing and a workaround before this goes to the RC phase.

As for the other items I listed, I think those could just be tested in the RC phase. In other words, my feeling is that User-Specific caching is the most pressing issue. The other items I listed I'm not too concerned about, and I don't _think_ we are going to see a bunch of problems with any of those. I just wanted to list them because there's a _chance_ we might find an issue in one of those. Not likely though.

raamdev:
> We may need to disable Memcached automatically for logged-in user caching

I'm fine with doing that for now, at least until this feature has been more thoroughly tested and we have more user feedback. It might be useful to show a one-time Dashboard message whenever a site owner enables both of those features explaining that Memcached / RAM caching will be disabled for Logged-In User caching due to reduced performance associated with the transient nature of a Logged-In User cache.

jaswrks:
@raamdev Memcached has been disabled for logged-in users now.

However, please take note of the following:

- Please close this PR (do not merge).
- Instead, merge `feature/879` https://github.com/websharks/comet-cache-pro/pull/315, which includes all of the changes here, and also some additional changes; i.e., proxy endpoint changes, and Memcached being disabled for logged-in users. Additional changes start here: https://github.com/websharks/comet-cache-pro/pull/315#pullrequestreview-35670628

PR #315, opened 2017-04-28
API Endpoints / Proxy Update

jaswrks:
See: https://github.com/websharks/comet-cache/issues/879

### Summary of Changes

- Use simple static text files on the CDN to check for the latest version.
- Checks for the latest version now occur over `http://` to avoid needing a fallback.
- Authentication for access to the latest zip file continues to come through cometcache.com. However, we now add an identifying header: `x-via-software: [internal method name]` to help us identify these in logs and in rate-limiting configurations in the future.
- Fallback URL changed from `proxy.websharks-inc.net` to `http://update-fallback.wpsharks.io/cc-proxy`, which is a new API endpoint fallback at wpsharks.io.
- Only try the fallback URL whenever there was an SSL-related error on the primary API endpoint; i.e., avoid retrying via the proxy whenever the problem is something other than an SSL issue.
- The API itself has been updated and now returns non-SSL download links using a new API endpoint at wpsharks.io: `dls.wpsharks.com/cdn-xar`. This avoids us needing to use the proxy endpoint to handle downloads too, on top of being a fallback for the initial connection. In short, the proxy is only used to handle authentication. From there, the response will include a download link that leads to `dls.wpsharks.io`, which delivers a file from the CDN over http://
- Polished the update methods and stats-related methods.


jaswrks:
@raamdev Please review when you have time. This PR includes changes from the work I did on Memcached, but you only need to review https://github.com/websharks/comet-cache-pro/pull/315/commits/89c92f5d5f8ba069c7c678e1a9823a0495be4de4 here. See also, summary of changes above.

raamdev (approved):
Woohoo! Fantastic! I just reviewed https://github.com/websharks/comet-cache-pro/pull/315/commits/89c92f5d5f8ba069c7c678e1a9823a0495be4de4 and it looks great. Thanks for including the code comments—those were helpful.

I'm marking the changes to https://github.com/websharks/comet-cache-pro/pull/315/commits/89c92f5d5f8ba069c7c678e1a9823a0495be4de4 approved, but I'm going to hold off merging this until I merge the PR for Memcached support—I'm still waiting for an answer to a question about further testing: https://github.com/websharks/comet-cache-pro/pull/314#issuecomment-298076755

jaswrks:
@raamdev This branch is ready to merge.

PR #316, opened 2017-05-26
[PR] Feature Request: Add Referrer-Policy HTTP Header to Whitelist

raamdev:
See: websharks/comet-cache#892


raamdev:
@jaswrks Any reason we can't provide a filter here in case a site owner wants to tweak the whitelist on their own?

jaswrks:
Nope. Sounds like a good idea to me too.

You can use something like:

```php
$headers = $this->applyWpFilters(GLOBAL_NS.'_cacheable_headers', $headers);
```

This will work too, 

```php
$headers = apply_filters(GLOBAL_NS.'_cacheable_headers', $headers);
```

and now that we are beyond the point of needing to support back compat. for ZenCache hooks, we should probably update the codebase. At some point we can convert all calls to `applyWpFilters()` to just `apply_filters()`. The same for `doWpAction()`, which can just be `do_action()`.



Beyond that, at some point we should do another review of the AC plugin hook system in Comet Cache, because now that WordPress is exposing the standard WP hook/filter library in the early phase, we shouldn't need that any longer.

raamdev:
@jaswrks Thanks! See https://github.com/websharks/comet-cache-pro/pull/316/commits/9c27afa63d4f8a5bb2f7373101ac4d9b57e08a95. This PR ready for a final review when you get a chance.

jaswrks:
Awesome. This looks great to me! 💯

jaswrks (approved):

PR #317, opened 2017-05-27
[PR] Refactor codebase to remove unnecessary applyWpFilters() and doWpAction()

raamdev:
See: websharks/comet-cache#893


raamdev:
@jaswrks How about [`applyFilters()`](https://github.com/websharks/comet-cache-pro/blob/170220/src/includes/traits/Shared/HookUtils.php#L212) and [`doAction()`](https://github.com/websharks/comet-cache-pro/blob/170220/src/includes/traits/Shared/HookUtils.php#L181)?

jaswrks:
@raamdev writes...
> @jaswrks How about applyFilters() and doAction()?

Awesome work on removing `doWpAction()` and `applyWpFilters()`. Everything else in that set of `HookUtils{}` represents our AC plugin hook system. Which is currently being used by some developers, so removing it will be a little trickier.

I think the following would work:

- Leave `HookUtils{}` where it is for now. Keeping that class long-term in case it's needed again in the future is not a bad idea either.
- Leave all existing calls to `applyFilters()`, `doAction()`, `addAction()`, etc. However, we should add new calls alongside them that use standard hook/filter functions in WP; i.e., deprecate the use of custom hook/filter methods in Comet Cache, in favor of core functions.
- Update [this example script](https://raw.githubusercontent.com/websharks/comet-cache-pro/dev/src/includes/templates/ac-plugin.txt) to use the WP core functions.

### New AC Plugin Example

```php
<?php
/**
 * Example AC (Advanced Cache) Plugin File.
 *
 * If implemented; this file should go in this special directory:
 *    `/wp-content/ac-plugins/my-ac-plugin.php`
 */
if (!defined('WPINC')) {
    exit('Do NOT access this file directly.');
}
function my_ac_version_salt_shaker($version_salt)
{
    if (mb_stripos($_SERVER['HTTP_USER_AGENT'], 'iphone') !== false) {
        $version_salt .= 'iphones'; // Give iPhones their own variation of the cache.
    } elseif (mb_stripos($_SERVER['HTTP_USER_AGENT'], 'android') !== false) {
        $version_salt .= 'androids'; // Androic variation.
    } else {
        $version_salt .= 'other'; // A default group.
    }
    return $version_salt;
}
add_filter('comet_cache_version_salt', 'my_ac_version_salt_shaker');
```

### Old AC Plugin Example

```php
<?php
/**
 * Example AC (Advanced Cache) Plugin File.
 *
 * If implemented; this file should go in this special directory:
 *    `/wp-content/ac-plugins/my-ac-plugin.php`
 */
if (!defined('WPINC')) {
    exit('Do NOT access this file directly.');
}
function my_ac_plugin() // Example plugin.
{
    $ac = $GLOBALS['comet_cache_advanced_cache']; // Comet Cache instance.
    $ac->addFilter('comet_cache_version_salt', 'my_ac_version_salt_shaker');
}
function my_ac_version_salt_shaker($version_salt)
{
    if (mb_stripos($_SERVER['HTTP_USER_AGENT'], 'iphone') !== false) {
        $version_salt .= 'iphones'; // Give iPhones their own variation of the cache.
    } elseif (mb_stripos($_SERVER['HTTP_USER_AGENT'], 'android') !== false) {
        $version_salt .= 'androids'; // Androic variation.
    } else {
        $version_salt .= 'other'; // A default group.
    }
    return $version_salt;
}
my_ac_plugin(); // Run this plugin.
```

---

The only change is that you can now call WP core functions in this early phase. You still need a special file in the `wp-content/ac-plugins` directory however.

jaswrks (approved):

PR #318, opened 2017-06-16
Bug Fix: Notice: id was called incorrectly. Product properties should…

vestaxpdx:
This fixes a PHP notice error when going through a checkout on Woocommerce. Since version 3.0 product properties should not be accessed directly. Instead we must use get_id() or get_data() etc...


jaswrks:
This looks great to me.

jaswrks (approved):

